PR #8, opened 2026-04-21
feat: POST /v1/context/build — context engine without inference

chazmaniandinkle:
## Summary

New endpoint that runs the CogOS context engine (thread parse, session resolve, compressor) and returns the constructed `ContextWindow` as JSON — without invoking inference (no Claude CLI, no `chat.response` bus event, no tool bridge).

Lets external callers use CogOS as the context authority while doing generation elsewhere. The immediate use case: an LM Studio generator plugin that calls this endpoint for context, then hands the constructed messages to the local LM Studio model for inference. In the framing of the research thesis — this is the EA/EFM split made concrete. The substrate decides what matters; the model generates.

## Contract

- **Method:** `POST /v1/context/build`
- **Request body:** same shape as `ChatCompletionRequest` (`messages`, `model`, `system_prompt`, `taa`, `tools`). `stream` and `tools` are accepted but have no effect on output.
- **Headers:** same as `/v1/chat/completions` (`X-Session-ID`, `X-Eidolon-ID`, `X-Origin`, `X-TAA-Profile`, `X-Session-Reset`, `X-OpenClaw-User-ID`/`-Name`, UCP packets).
- **Response** (200 JSON):

```json
{
  "session_id": "http",
  "claude_session_id": "49580423-...",
  "strategy": "resume|fresh|fallback",
  "total_tokens": 420,
  "system_prompt": "...",
  "prompt": "...",
  "taa": {
    "profile": "openclaw",
    "enabled": true,
    "total_tokens": 1024,
    "coherence": 0.83,
    "tiers": {"tier1": 256, "tier2": 128, "tier3": 512, "tier4": 128}
  }
}
```

## Errors

- **400** `invalid_request` — malformed JSON, invalid UCP headers, no user message
- **405** `invalid_request` — non-POST method
- **503** `unavailable` — context engine not initialized
- **500** `context_build_failed` — context engine returned an error during `Build`

## Side effects

By design, this endpoint is quieter than `/v1/chat/completions`:

| Side effect | `/v1/chat/completions` | `/v1/context/build` |
|---|---|---|
| `lifecycle.GetOrCreate` | ✓ | — |
| `appendToThread("user", ...)` | ✓ | — |
| `busChat.emitRequest` (chat.request) | ✓ | — |
| `sessionMgr.Rotate` (inside `ContextEngine.Build`) | ✓ | ✓ |
| Claude CLI inference | ✓ | — |
| `chat.response` bus event | ✓ | — |
| Tool bridge continuation | ✓ | — |

Rationale for the lighter footprint: external callers using this endpoint don't want to pollute session state with speculative context-build calls. If full-side-effect parity is needed later, gate behind `?commit=1`.

## Tests

New `serve_context_build_test.go` — 6 cases:

- `TestHandleContextBuild_MalformedJSON` → 400
- `TestHandleContextBuild_NoUserMessage` → 400
- `TestHandleContextBuild_NoContextEngine` → 503
- `TestHandleContextBuild_MethodNotAllowed` → 405
- `TestHandleContextBuild_BasicRequest` → 200 with documented shape
- `TestHandleContextBuild_ExplicitSessionID` → session_id passthrough

Full `go test ./...` passes (8.1s, no regressions).

## Live smoke test

Deployed to a local kernel, LAN-accessible:

```bash
$ curl -s -X POST http://localhost:6931/v1/context/build \
    -H "Content-Type: application/json" \
    -d '{"messages":[{"role":"user","content":"What is 2+2?"}]}'
{
  "session_id": "http",
  "strategy": "fresh",
  "total_tokens": 15,
  "system_prompt": "",
  "prompt": "## Earlier Context\n[user] What is 2+2?\n\n---\n\nWhat is 2+2?"
}
```

## Follow-ups (not in this PR)

- `?commit=1` flag for full-side-effect parity with chat completions, if the use case emerges
- Shared helper extracted from both handlers (currently the two endpoints duplicate the UCP-parse / prompt-flatten / session-ID-derivation preamble — fine for a spike, worth revisiting once both surfaces have settled)
- Bus enrichment block in the response (bus_id, seq, hash) — deferred since bus emission doesn't fire here

## Test plan

- [ ] Review endpoint contract + side-effect policy
- [ ] `go test ./...` runs green on CI
- [ ] Smoke test: `curl -X POST /v1/context/build` returns a populated `ContextWindow`-shaped payload


chazmaniandinkle (commented):
## Codex pre-merge review + amendment

Ran a read-only Codex review (gpt-5.3-codex, high reasoning) against the diff. Two findings surfaced; amendment commit `65371e6` addresses one, the other is deferred as out-of-scope.

### Addressed in `65371e6`

**Nil-kernel topology gap.** `workspaceRoot` was only sourced from `s.kernel.Root()`, so when `kernel` is nil but `contextEngine` is populated (the test fixture topology, and observer/edge-caller scenarios), UCP/TAA paths would return spurious 400s or crash on missing profile YAML. Referenced: `serve_context_build.go:71, 77, 197` and test fixture at `serve_context_build_test.go:19`.

The fix:
- UCP parsing is skipped gracefully when `workspaceRoot == ""` (log line fires if UCP headers were supplied so the degradation is visible)
- TAA diagnostics downgrade to `{enabled: true}` with empty tiers instead of crashing on `ConstructContextStateWithProfile`

Added `TestHandleContextBuild_NilKernelWithUCPHeaders` guarding the regression. All 7 `TestHandleContextBuild_*` tests pass, full `go test ./...` green.

### Deferred — not in this PR

**Tests bypass the real mux.** Handler is called directly in tests rather than through `http.NewServeMux` with the method pattern registered at `serve.go:226`; this misses mux-level method handling and middleware, so 405 behavior and middleware wiring aren't exercised at the HTTP boundary.

Fixing this is a project-wide testing-style decision (move all handler tests to an `httptest.NewServer`-backed fixture? introduce a router-aware helper?), not a mechanical change appropriate for this PR's scope. Tracked separately.

### Verdict

Codex returned "needs changes"; the one in-scope change is addressed.

PR #58, opened 2026-04-28
Fix Codex app bundle resolution

### LINKED ISSUE #57: Codex provider unavailable under LaunchAgent PATH; model=codex falls back to Claude
## Summary

Requests to the OpenAI-compatible `/v1/chat/completions` endpoint with `model: "codex"` currently appear to succeed with `model:"codex"` in the HTTP response, but the kernel selects `claude-code` because the `codex` provider is unavailable in the launchd daemon environment.

## Evidence

Live probe on 2026-04-28:

- Request: `POST /v1/chat/completions` with `model:"codex"`
- HTTP response: `"model":"codex"`
- `/v1/debug/last`: selected provider was `claude-code`
- Process sampler observed child process: `claude -p --model sonnet ...`
- `/v1/providers` reported:

```json
{
  "name": "codex",
  "type": "codex",
  "available": false
}
```

## Likely Cause

`CodexProvider.Available()` uses `exec.LookPath(p.binary)`. When no `endpoint` is configured, `p.binary` defaults to `codex`.

In the user shell, Codex resolves to:

```text
/Applications/Codex.app/Contents/Resources/codex
```

But the `com.cogos.kernel` LaunchAgent PATH does not include `/Applications/Codex.app/Contents/Resources`, so the daemon cannot resolve `codex` and marks the provider unavailable.

## Impact

- `model:"codex"` silently falls back to `claude-code`.
- The OpenAI-compatible response still reports `model:"codex"`, which can mislead clients/operators.
- The Codex provider code exists and is registered, but is not actually used in this environment.

## Proposed Fix

Make the Codex provider robust to the Codex desktop app install path, for example:

- Resolve `codex` through PATH first.
- If not found and running on macOS, fall back to known app bundle locations such as `/Applications/Codex.app/Contents/Resources/codex`.
- Reuse the resolved binary path in `Ping`, `Complete`, and `Stream`.
- Add focused tests for PATH failure + app bundle fallback behavior.

A local config workaround is to set:

```yaml
codex:
  type: codex
  endpoint: /Applications/Codex.app/Contents/Resources/codex
  model: gpt-5.4
  timeout: 120
  options:
    effort: high
    sandbox: read-only
```

But the provider should handle the common Codex.app installation path without requiring every LaunchAgent environment to mirror the interactive shell PATH.

chazmaniandinkle:
## Summary

Fixes #57.

- Resolves the Codex CLI through PATH first for the default provider binary.
- Falls back to known macOS Codex.app bundle locations when launchd PATH cannot find `codex`.
- Preserves explicit `endpoint` overrides by avoiding app-bundle fallback for configured binary names or paths.
- Uses the same resolver for availability, ping, completion, and streaming.

## Validation

- `go test ./internal/engine -run 'TestCodex'`\n- `go test ./internal/engine`

### DIFF
--- a/internal/engine/provider_codex.go
+++ b/internal/engine/provider_codex.go
@@ -12,20 +12,32 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"os"
 	"os/exec"
+	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 )
 
+var (
+	codexLookPath = exec.LookPath
+	codexStat     = os.Stat
+	codexGOOS     = runtime.GOOS
+)
+
+const codexAppBundleBinary = "/Applications/Codex.app/Contents/Resources/codex"
+
 // CodexProvider implements Provider by spawning codex exec processes.
 type CodexProvider struct {
-	name    string
-	model   string // "gpt-5.4", "gpt-5.3-codex-spark", etc.
-	effort  string // "xhigh", "high", "medium", "low"
-	sandbox string // "read-only", "workspace-write", "danger-full-access"
-	timeout time.Duration
-	binary  string // path to codex binary (default: "codex")
-	workDir string // working directory for codex exec
+	name          string
+	model         string // "gpt-5.4", "gpt-5.3-codex-spark", etc.
+	effort        string // "xhigh", "high", "medium", "low"
+	sandbox       string // "read-only", "workspace-write", "danger-full-access"
+	timeout       time.Duration
+	binary        string // path to codex binary (default: "codex")
+	defaultBinary bool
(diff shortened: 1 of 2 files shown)


chazmaniandinkle:
Validation after subagent implementation:

- Reviewed the diff for issue #57.
- Added an executable-bit guard for the Codex.app fallback path and a regression test for non-executable fallback files.
- Amended the single commit with corrected author metadata and no co-author trailer.
- Re-ran local tests:
  - `go test -count=1 ./internal/engine -run 'TestCodex'`
  - `go test -count=1 ./internal/engine`
- GitHub checks are green: build, changelog, lint, test, and e2e.

Issue #10, opened 2026-04-21
bug: EmitLedgerEvent bypasses hash chaining and writes to flat .cog/ledger/events.jsonl


chazmaniandinkle:
## Context

Surfaced during the design work for the ledger read API (PR against `feat/ledger-api`, Agent L's design). `EmitLedgerEvent` in [`internal/engine/mcp_stubs.go:229-252`](https://github.com/cogos-dev/cogos/blob/main/internal/engine/mcp_stubs.go#L229-L252) is the write path used by the `cog_emit_event` MCP tool ([`mcp_server.go:820`](https://github.com/cogos-dev/cogos/blob/main/internal/engine/mcp_server.go#L820)).

## Current behavior

`EmitLedgerEvent` opens `.cog/ledger/events.jsonl` at the workspace root — a **flat, non-hash-chained file** — and appends a bare `map[string]any` serialization. It does not:

- Use `AppendEvent`
- Participate in the per-session hash chain
- Get a sequence number
- Land in the per-session directory (`.cog/ledger/<sessionID>/events.jsonl`)

## Impact

Events emitted via the `cog_emit_event` MCP tool are **orphaned**. After the ledger read API lands (`cog_read_ledger` / `GET /v1/ledger`), these events will not be visible — the query scans per-session dirs, not the flat root file. The tamper-evident chain doesn't include them either.

## Expected behavior

Route through `AppendEvent(workspaceRoot, sessionID, envelope)` using the current session ID available from `process.SessionID()` (or an equivalent). The envelope should be a properly constructed `EventEnvelope` with a `EventPayload` containing type, timestamp, session_id, and the supplied event data.

## Suggested fix shape

1. Refactor `EmitLedgerEvent` to accept or derive a session ID, build an `EventEnvelope`, and call `AppendEvent`.
2. Update `toolEmitEvent` at `mcp_server.go:820` to pass the correct session context (if the signature changes).
3. Delete the flat `.cog/ledger/events.jsonl` path or document it as legacy — it should no longer be written.
4. Backfill test coverage: an MCP-emitted event should be readable via the new `cog_read_ledger` tool with `session_id=<current>`.

## References

- `internal/engine/ledger.go:142` — `AppendEvent` (correct write path)
- `internal/engine/ledger.go:40-59` — `EventEnvelope` / `EventPayload` / `EventMetadata` types
- `internal/engine/mcp_stubs.go:229-252` — the broken function
[...]

PR #106, opened 2026-05-01
fix(engine): pin Ollama model in VRAM via keep_alive=-1

chazmaniandinkle:
## Summary

Sets `keep_alive: -1` on every Ollama `/api/chat` request the kernel issues — both via `OllamaProvider` (metabolic loop) and `AgentHarness` (`cog_dispatch_to_harness` MCP path) — so the local model stays resident between cycles.

## Context

Every Ollama request from the kernel today omits `keep_alive`, so Ollama applies its default 5-minute window. The metabolic loop's adaptive cadence and `cog_dispatch_to_harness` arrivals are both bursty: any wake past 5 min pays a full cold-load (~110s for `gemma4:e4b`). That's the user-visible "every message takes forever to start" symptom.

## Changes

**Commit 1** — `internal/engine/provider_ollama.go` (`OllamaProvider`, used by the metabolic loop and the local-harness `runCycle`):
- New top-level `KeepAlive any` field on `ollamaChatRequest` (`keep_alive,omitempty`).
- `buildOllamaRequest` defaults it to `-1`.
- `TestBuildOllamaRequestKeepAlivePinsModel` asserts the marshaled body contains `"keep_alive":-1`.

**Commit 2** — `agent_harness.go` (`AgentHarness`, used by the `cog_dispatch_to_harness` MCP path via `HarnessDispatcher`):
- `KeepAlive any` added to `agentChatRequest`.
- Set to `-1` at all three construction sites: `Assess`, `Execute`, `ExecuteScoped`.
- On the OpenAI-compat `/v1/chat/completions` path the field is silently dropped (`chatCompletionOpenAI` translates to its own struct that doesn't forward it) — desired behavior; LM Studio and the like don't honor `keep_alive` anyway.
- `TestAgentHarness_Assess_KeepAlivePinsModel` is a wire-level test: decodes the raw POST body and asserts `keep_alive` parses as `-1`. A refactor that drops the field or changes its JSON tag fails this test rather than silently regressing.

## Correction to prior PR description

The first revision of this PR claimed `agent_harness.go` was orphaned. That was wrong — `agent_dispatch.go` holds a live `*AgentHarness` and is the MCP dispatch path users actually hit. Commit 2 fixes that gap.

## Out of scope (not in this PR)

The actual *stacking* symptom (multiple model copies resident when several user messages arrive close together) has a separate root cause: `LocalHarnessController.runCycle` and `LocalHarnessController.DispatchToHarness` both build their own provider with no shared serialization, and `DispatchToHarness` itself fans out `req.N` goroutines per call. Tracked in #107.

## Test plan

- [x] `go vet ./...`
- [x] `go build -o cogos ./cmd/cogos`
- [x] `go build -tags mcpserver -o cogos-mcp ./cmd/cogos`
- [x] `go test -race -count=1 ./...` (root + `internal/engine` packages)
- [ ] Smoke test against a real Ollama: a second request issued >5 min after the first should not show `load_duration` > 0 in the response metrics.

### DIFF
--- a/agent_harness.go
+++ b/agent_harness.go
@@ -128,6 +128,13 @@ type agentChatRequest struct {
 	Think    bool                   `json:"think"`             // explicit thinking control
 	Format   string                 `json:"format,omitempty"`  // "json" for structured output
 	Options  map[string]interface{} `json:"options,omitempty"` // Ollama model options (num_ctx, temperature, etc.)
+	// KeepAlive controls how long Ollama keeps the model resident after the
+	// request completes. -1 keeps the model loaded indefinitely, 0 unloads
+	// immediately. Set to -1 by callers so dispatch cycles spaced past
+	// Ollama's 5-minute default keep-alive don't pay cold-start. Dropped
+	// on the OpenAI-compat path (chatCompletionOpenAI translates to its
+	// own struct that does not forward this field).
+	KeepAlive any `json:"keep_alive,omitempty"`
 }
 
 // agentChatMessage is a single message in the conversation.
@@ -249,12 +256,13 @@ func (h *AgentHarness) Assess(ctx context.Context, systemPrompt, observation str
 	}
 
 	req := agentChatRequest{
-		Model:    h.model,
-		Messages: messages,
-		Stream:   false,
-		Think:    false, // disable thinking — we want clean JSON output
-		Format:   "json",
-		Options:  map[string]interface{}{"num_ctx": 8192}, // 8K context is plenty for assessment
+		Model:     h.model,
+		Messages:  messages,
+		Stream:    false,
+		Think:     false, // disable thinking — we want clean JSON output
+		Format:    "json",
+		Options:   map[string]interface{}{"num_ctx": 8192}, // 8K context is plenty for assessment
+		KeepAlive: -1,                                      // pin model in VRAM between cycles
 	}
 
 	resp, err := h.chatCompletion(ctx, req)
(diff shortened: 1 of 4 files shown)


chazmaniandinkle:
Filed chazmaniandinkle/cogos#10 as follow-up: extends wire-body test coverage to Execute/ExecuteScoped (table-driven, all three paths in one helper) and documents the manual VRAM recovery step. Stacked on this PR's branch so the diff is only the follow-up work; will rebase onto main after this merges.

chazmaniandinkle:
## Code review

**Recommendation: approve and merge.**

Targeted performance fix: adds `keep_alive: -1` on every Ollama `/api/chat` request from both `OllamaProvider` (metabolic loop) and `AgentHarness` (`cog_dispatch_to_harness` MCP path). Replaces ~110s cold-load on bursty wakes past Ollama's 5-min default.

### Strengths
- Tight scope (+100/-24, 4 files; no drive-by refactors).
- Wire-level test (`TestAgentHarness_Assess_KeepAlivePinsModel`) decodes the actual outbound HTTP body — a refactor that drops the field will fail loudly rather than silently regress.
- PR body is exemplary: explicit self-correction of an earlier wrong claim about `agent_harness.go` being orphaned, and a clean out-of-scope boundary referencing #107.

### Optional polish before merge
- Negative test on the OpenAI-compat path asserting the body does **not** contain `keep_alive` — cheap insurance against future translation-layer changes.
- `KeepAlive any` + `omitempty` is silent if a future caller forgets to set it (zero value `nil` drops the field). Consider routing struct construction through a helper (`newAgentChatRequest(...)` pre-filling `-1`) so future sites can't omit it.

### Tracked elsewhere (not blocking)
- Wire-level coverage for `Execute`/`ExecuteScoped`: follow-up #10.
- Stacking symptom under concurrent dispatch: #107.

PR #111, opened 2026-05-01
fix(bus): wire BusEventBroker to /v1/bus/{id}/stream SSE endpoint

### LINKED ISSUE #51: Bus events never reach /v1/events/stream; BusEventBroker has no HTTP route
## Summary

In cogos-v3, bus events (appends via `/v1/bus/send`) propagate to an internal `BusEventBroker` that has **no HTTP route registered**. The `/v1/events/stream` SSE handler only emits ledger events. Clients using `cog bus watch` (which filters for `sseEvent.Type == "bus.event"` frames) sit silent indefinitely — the server structurally never emits such frames.

This silently broke push-style bus subscription for downstream consumers (hooks, watchers, dashboards).

## Evidence

- `internal/engine/serve_events.go` — ledger SSE handler, around line 134. Emits only `event: ledger.appended` frames.
- `internal/engine/bus_stream.go` — contains the orphaned `BusEventBroker`. Header comment explicitly states *\"No HTTP route is registered in this phase.\"*
- `internal/engine/serve_bus.go` — bus routes. No `/stream` endpoint registered.
- Client-side: the `cog bus watch` CLI (at `.cog/bus_transport_sse.go` in the consumer workspace) connects to `/v1/events/stream` and filters on `sseEvent.Type == \"bus.event\"`. Server never emits this type under v3. Silent fall-through, no error, no exit.

Empirical probe: 9 seconds of unfiltered SSE on `/v1/events/stream` during a period when the kernel was firing 10k+ `bus_cycle_trace` events/minute produced zero `bus.event` frames — only the initial `connected` handshake. Polling `GET /v1/bus/{id}/events?after_seq=N` returns every event correctly with full hash chain.

## Impact

- `cog bus watch` is structurally silent under v3. No error surfaces; the tool appears to run fine.
- Any monitor, hook, or dashboard assuming push-style bus subscription receives zero events.
- Polling via `/v1/bus/{id}/events?after_seq=N` is the only working path.
- Observed during RFC-003 (client-side peer-awareness) work on 2026-04-23.

## Proposed fixes (either resolves it)

**Option A: Register a dedicated bus-stream route.**
Mount a handler at e.g. `GET /v1/bus/{id}/stream` that subscribes to `BusEventBroker`. Update `cog bus watch` client to use that route. Preserves the broker separation (ledger vs bus).

**Option B: Bridge BusBlock appends into the ledger EventBroker.**
Forward bus appends through the existing `EventBroker` as `event: ledger.appended` frames with `source=bus:<id>`. Client filter adjusts to match. Simpler mechanically (one broker, one route) but blurs the semantic distinction between ledger and bus events.

## Ancillary findings (separate but discovered together)

- `/v1/events/stream` accepts but silently ignores unknown query params (e.g. `bus_id=<id>`). Client appends it; server does not validate. Adding strict validation would surface this class of client-server mismatch.
- `.cog/.state/buses/registry.json = []` despite 200+ bus directories on disk. Rehydration decoupled from directory presence. Unrelated to this SSE gap but discovered in the same investigation; may cause cross-restart continuity problems for RFC-003's per-channel topics.

## Workaround (for downstream consumers)

Until fixed, use polling:

\`\`\`python
while True:
    evs = fetch(f'/v1/bus/{bus}/events?after_seq={seq}')
    for e in evs:
        seq = e['seq']
        # process event
    time.sleep(2)
\`\`\`

2s cadence works for peer-awareness; 500ms for interactive.

## Files most relevant

- `internal/engine/serve_events.go` (ledger SSE handler)
- `internal/engine/bus_stream.go` (orphaned broker)
- `internal/engine/serve_bus.go` (bus routes)

chazmaniandinkle:
Closes #51.

## Summary

\`BusEventBroker\` was fully functional and received published events from every \`POST /v1/bus/send\` call but had no HTTP route registered. The \`bus_watch.go\` client was connecting to \`/v1/events/stream\` (the ledger SSE endpoint) which only emits \`ledger.appended\` frames, never bus-domain events.

The header comment in \`bus_stream.go\` itself stated: \"No HTTP route is registered in this phase.\"

- Adds \`handleBusStream\` to \`serve_bus.go\` (~100 lines): subscribes to \`BusEventBroker\`, optionally replays disk history on connect (gated by \`?no_replay=1\`), then streams live events in the \`{id, type, timestamp, data}\` envelope shape that \`bus_watch.go\` already expects.
- Adds \`busSSEEnvelope\` wire type matching the existing \`busEventEnvelope\` in \`bus_watch.go\`.
- Mounted under the existing \`/v1/bus/\` catch-all as \`case \"stream\"\` — no new top-level route registration needed.
- Updates \`bus_watch.go runLive\` from \`/v1/events/stream?bus_id=<id>\` to \`/v1/bus/<id>/stream\`, propagates \`no_replay=1\` when \`--no-replay\` is set.
- \`TestBusStreamSSEEndToEnd\` opens a bus, subscribes with \`no_replay=1\`, sends two events via \`POST /v1/bus/send\`, asserts both arrive on the stream with non-empty IDs and non-nil data.

\`make test\` passes (full engine suite green; pre-existing root-package failure unrelated).

## Strategic note

This unblocks SSE-based dashboard updates. Any consumer of \`GET /v1/bus/{id}/stream\` now gets true push delivery. The \`BusEventBroker\`'s per-bus fan-out (with backpressure drop and idle reaper) is already production-quality — the missing piece was only the HTTP surface.

## Notes for review

- Subscription happens before replay so no events are lost in the seam between snapshot and live subscription.
- Replay reads ALL historical events on connect via \`ReadEvents(busID)\`. For long-lived buses with thousands of events this is O(N); consider a \`?since=<seq>\` parameter as follow-up. Not a blocker.
- The \`replay_\` prefix on replay event IDs is a string convention; could be a sentinel const.
- Heartbeat at 30s (\`busSSEHeartbeatInterval\`); per-frame write deadline at 10s.

## Test plan

- [x] \`make test\` passes
- [x] Regression test \`TestBusStreamSSEEndToEnd\` covers the path
- [ ] Manual: \`curl -N localhost:6931/v1/bus/<id>/stream\` and emit an event

### DIFF
--- a/bus_watch.go
+++ b/bus_watch.go
@@ -42,9 +42,12 @@ func (w *BusWatcher) Run(ctx context.Context) error {
 
 // runLive connects to the kernel SSE endpoint and streams events.
 func (w *BusWatcher) runLive(ctx context.Context) error {
-	url := fmt.Sprintf("http://%s/v1/events/stream?bus_id=%s", w.kernelAddr, w.busID)
+	rawURL := fmt.Sprintf("http://%s/v1/bus/%s/stream", w.kernelAddr, w.busID)
+	if w.filter.NoReplay {
+		rawURL += "?no_replay=1"
+	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, rawURL, nil)
 	if err != nil {
 		return fmt.Errorf("create request: %w", err)
 	}
--- a/internal/engine/serve_bus.go
+++ b/internal/engine/serve_bus.go
@@ -14,6 +14,7 @@
 //	GET    /v1/bus/{bus_id}/events            — per-bus events, filtered
 //	GET    /v1/bus/{bus_id}/events/{seq}      — single event by seq
 //	GET    /v1/bus/{bus_id}/stats             — bus statistics
+//	GET    /v1/bus/{bus_id}/stream            — live SSE stream for a bus
 //	GET    /v1/bus/consumers                  — list consumer cursors (ADR-061)
 //	DELETE /v1/bus/consumers/{consumer_id}    — remove a consumer cursor
 package engine
@@ -25,6 +26,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/cogos-dev/cogos/pkg/cogfield"
 )
@@ -320,8 +322,10 @@ func (s *Server) handleBusRoute(w http.ResponseWriter, r *http.Request) {
 		}
 	case "stats":
 		s.handleBusStats(w, r, busID)
+	case "stream":
+		s.handleBusStream(w, r, busID)
 	default:
-		http.Error(w, `{"error":"expected /v1/bus/{bus_id}/events or /v1/bus/{bus_id}/stats"}`, http.StatusBadRequest)
+		http.Error(w, `{"error":"expected /v1/bus/{bus_id}/events, /v1/bus/{bus_id}/stats, or /v1/bus/{bus_id}/stream"}`, http.StatusBadRequest)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


chazmaniandinkle:
## Code review

**Recommendation: ready-once-marked** (after the test fixes below).

Closes #51 by registering `GET /v1/bus/{bus_id}/stream` as the missing HTTP surface over the already-functional `BusEventBroker`. Subscribes before replay (correct ordering), emits a connected handshake, replays disk history (skippable via `?no_replay=1`), then streams live events with 30s heartbeats and 10s per-write deadlines. Implementation is small, well-scoped, and matches Option A from the issue.

### Strengths
- Subscribe-before-replay correctly closes the snapshot/live seam (no lost events; possible duplicates only — acceptable).
- Reuses existing `case` dispatch under `/v1/bus/`; no new top-level routing.
- Honors broker capacity (`Subscribe` returns false → 503), uses `consumer_id` dedup, `TouchWrite` keeps the idle reaper from killing active consumers during long replays.
- SSE headers correct; heartbeat uses comment frames (EventSource ignores them).

### Pre-merge fixes (small)
- Drop the `time.Sleep(50ms)` in the regression test — flake hazard. Poll `broker.SubscriberCount(busID) == 1` instead.
- Add a disconnect-cleanup assertion: `cancel()` should drop `SubscriberCount` back to 0.

### Hardening (file as follow-up issues, not blocking)
1. **Replay backpressure / write-deadline race.** During replay, every `fmt.Fprintf` resets a 10s deadline but only flushes once at the end. A slow client with thousands of replay events accumulates in `http.ResponseWriter` buffer; broker simultaneously fans live events on `ch` (drop-on-full). Consider flushing every N frames during replay; cap with `?since=<seq>` (PR description anticipates this).
2. **Drop semantics not surfaced.** `Publish` silently drops on a full channel (buf 64). At minimum log at debug; ideally emit an `event: lag` frame so clients fall back to polling — important since the watcher CLI now relies on this exclusively.
3. **Unbounded replay.** `ReadEvents(busID)` returns ALL history. O(N) memory + I/O hit per connect, DoS vector. `?since=<seq>` should land before this is heavily used.
4. **`replay_` ID prefix** could collide with future legitimate hashes. Cheap to fix: separate envelope field like `"replay": true`.

### Minor
- `Connection: keep-alive` is hop-by-hop, set automatically by net/http, ignored under HTTP/2 — harmless noise.
- Validate the bus exists before subscribing (today an unknown busID succeeds and silently never delivers).
- Other test gaps worth filling: heartbeat emission, `?no_replay=0` path replaying prior events, broker-at-capacity 503, `consumer_id` dedup eviction.

Issue #90, opened 2026-04-29
field: split inbox attention boosts out of the chat-read salience score


chazmaniandinkle:
## Symptom

After landing the foveated-assembler `MaxFovealDocs=10, SalienceFloor=0.3` cap (issue separate), the top admitted CogDocs in chat-path debug snapshots have **`relevance: 0.00, reason: 'both'`** for every entry. They're admitted purely on field salience. The top admitted doc on a typical chat turn:

\`\`\`
sal=0.876  cog://mem/semantic/inbox/conversations/chatgpt/chatgpt-2025-09-25-langgraph-openrouter-agent.cog.md
sal=0.702  cog://mem/semantic/inbox/conversations/chatgpt/.../fire-emblem-fates-emulator.cog.md
sal=0.702  cog://mem/semantic/inbox/conversations/chatgpt/.../pizza-vrchat-combo.cog.md
…
\`\`\`

These are unprocessed inbox entries (ChatGPT exports, Discord links). They are not relevant to anything the user is currently asking the kernel about.

## Root cause

`internal/engine/field.go:32–36` defines two flat-add boosts that `applyInboxBoosts` (`field.go:326–339`) adds to every file under any `/inbox/` path:

\`\`\`go
inboxRawBoost      = 0.5  // status: raw
inboxEnrichedBoost = 0.2  // status: enriched
\`\`\`

`field.go:122` calls `applyInboxBoosts(fresh)` after every full salience recomputation, so the same boosted scores feed both the **observer loop** (which is what they're for — drawing attention to unprocessed items) and the **chat-path foveated assembler** (which doesn't want this — it wants doc relevance to the user's current query).

With the langgraph file at `+0.7` (raw + recency), and ~4,800 enriched-status files at `+0.2`, the chat assembler sees an inbox-dominated field and admits whichever bumped-up files clear the salience floor.

## Fix

Split `Score()` into two views, populated in the same scan:

- `Score(path) → float` — base score, no inbox boosts. Used by the chat-path foveated assembler (`context_assembly.go:303`) and the foveated endpoint (`serve_foveated.go:160`).
- `ObserverScore(path) → float` — base + inbox boosts. Used by the observer loop only.

Sketch:

\`\`\`go
type AttentionalField struct {
    // ...
    base     map[string]float64 // no boosts
    observer map[string]float64 // base + inbox boosts
}

func (f *AttentionalField) Score(path string) float64        { return f.base[path] }
[...]

Issue #88, opened 2026-04-29
foveated-assembler: gating knobs + kernel-agent alias + settings endpoints (Spike 2/3 baseline)


chazmaniandinkle:
## Symptom

Dashboard chat path admits **416 CogDocs / 27,683 tokens** for a 4-token user query (32K budget eaten alive, 1 token of headroom, **zero room for chat history**). Separately, `model: kernel-agent` does not exist as a routing alias; dashboard chat lands on the `claude-code` subprocess (Anthropic cloud) while `cog dispatch` hits Ollama. The two surfaces don't share a harness.

## Root cause

Two findings from forensic audits at `.cog/scratch/audit-dashboard-context/REPORT.md` and `.cog/scratch/audit-inference-paths/REPORT.md`:

1. **`internal/engine/context_assembly.go:292–327`** — chat-path admission predicate is `if relevance <= 0 && salience <= 0 { continue }`. No floor, no `maxDocs` cap. Sibling `serve_foveated.go:189–195` (Claude Code hook) caps at `maxDocs := 10, tier4Budget := 4000`. Same field, different gating.

2. **`internal/engine/serve.go:722–754`** — model-alias switch has cases for `claude`, `codex`, `ollama` but no `kernel-agent`. The dashboard `dashboard.html:1019` defaults to `model: 'local'`, which falls through to the workspace `providers.yaml` `process_state_routing.active = claude-code`.

## Validated fix (binary-only, needs to land in source)

A field-tested binary patch produced **94% token reduction** (28K → 1.6K) and confirmed `provider.selected = 'ollama'` on `model: 'kernel-agent'`. The patch:

- `config.go`: add `MaxFovealDocs int` (default 10), `SalienceFloor float64` (default 0.3); RWMutex-protected `ContextGating()` / `SetContextGating()` accessors; YAML keys `max_foveal_docs:` and `salience_floor:`.
- `context_assembly.go`: in both TRM and keyword branches — read knobs at request time, apply salience floor before admission, truncate to `MaxFovealDocs` after sort.
- `serve.go`:
  - new model alias case: `case \"ollama\", \"kernel-agent\": creq.Metadata.PreferProvider = \"ollama\"`
  - new endpoints `GET /v1/settings/context` (returns current knobs) and `PATCH /v1/settings/context` (hot-update)
  - `handleDebugContext` decorates the response with `gating: {max_foveal_docs, salience_floor}` so dashboards can show what's active
- `dashboard.html`:
[...]

PR #112, opened 2026-05-01
feat(context): default_budget config + exclude_globs path filter + per-request budget header

### LINKED ISSUE #77: Context assembly injects unbounded CogDoc manifests; no per-client budget or doc-count cap
## Problem

The dashboard chat at :6931 receives a large CogDoc manifest per request
(~30K tokens observed). There is no way for the operator to configure the
budget, cap the number of injected docs, exclude path globs, or apply
different policies for different clients (dashboard chat vs. Claude Code
MCP hook vs. `cog_assemble_context` MCP tool).

The manifest is also weighted in a way that lets bulk-ingested inbox
material crowd out query-relevant docs. Any workspace with many
`/inbox/` items in `status: raw` will see them cluster at the top of the
ranking regardless of query relevance, because the inbox boost is applied
unconditionally and exceeds typical base-salience differences.

## Mechanism

`Process.AssembleContext()` (`internal/engine/context_assembly.go:154`) is the
single entry point. All three call sites pass `budget=0`:

- `serve.go:764` — OpenAI-compat `/v1/chat/completions`
- `serve_anthropic.go:171` — Anthropic-compat endpoint
- `mcp_server.go:891` — `cog_assemble_context` MCP tool (defaults to 50000)

Budget=0 falls through to `budget = 32768` at `context_assembly.go:200`. The
flex budget for docs+history is then `budget − output_reserve − nucleus −
client_system − current_msg`, ≈25–28K tokens.

`evictForBudgetModeWithEstimator()` (`context_assembly.go:471`) fills until
the budget runs out. **There is no per-doc count cap.**

`field.go:32` adds `inboxRawBoost = 0.5` to every `/inbox/` doc with
`status: raw`, and `0.2` for `enriched`. With empty/short queries the
keyword+salience fallback (`context_assembly.go:292–323`) ranks by
`relevance*2 + salience`, so the inbox boost dominates and pulls
unrelated inbox items into the foveal set ahead of query-matching docs.

## What's missing

`internal/engine/config.go` exposes only `output_reserve` and
`salience_days_window`. There is no config for:

- Total context budget (and per-endpoint overrides)
- Maximum number of foveal docs to include
- Path glob include/exclude lists
- Manifest mode on/off (currently hard-coded `true` at every call site)
- Inbox boost values
- Per-client policies (dashboard chat ≠ Claude Code hook ≠ MCP tool)

## Proposal

1. **Config keys** in `kernel.yaml` under a new `v3.context` block:
   - `default_budget` (default 32768)
   - `max_foveal_docs` (default e.g. 8)
   - `inbox_raw_boost`, `inbox_enriched_boost` (lift constants out of `field.go`)
   - `exclude_globs` to opt paths out of foveal injection until they're
     promoted out of inbox status
   - Per-client overrides keyed by route or User-Agent:
     ```yaml
     v3:
       context:
         clients:
           dashboard: { budget: 8192, max_foveal_docs: 5 }
           claude_code: { budget: 16384, max_foveal_docs: 10 }
           mcp_tool: { budget: 50000 }
     ```
2. **Pass-through**: a request-level override (e.g. `X-Cogos-Context-Budget`
   header on the OpenAI/Anthropic-compat endpoints, or a `budget` field in
   `cog_assemble_context`) so an agent can manage its own context actively
   — supporting future tools that ask the kernel for \"give me less\" or
   \"give me focused on X.\"
3. **Hard cap**: enforce `max_foveal_docs` in
   `evictForBudgetModeWithEstimator` in addition to the byte budget.
4. **Diagnostic endpoint / event**: the existing `context.assembly` event
   (`context_assembly.go:349`) is good — surface it on a `/v1/context/last`
   debug route so the operator can see exactly what got injected and why.

## Acceptance

- [ ] `kernel.yaml` keys above are honored
- [ ] Dashboard chat at :6931 produces ≤ configured budget
- [ ] An agent calling `cog_assemble_context` can pass its own budget
      and have it respected
- [ ] Inbox glob can be excluded via config without code change
- [ ] Per-client policies tested for dashboard, Claude Code, MCP tool

## Files

- `internal/engine/config.go:101` — config struct
- `internal/engine/context_assembly.go:154,200,471` — assembler + eviction
- `internal/engine/field.go:32` — inbox boost constants
- `internal/engine/serve.go:764`, `serve_anthropic.go:171`, `mcp_server.go:891` — call sites

chazmaniandinkle:
Closes #77.

## Summary

Issue #77 had three sub-problems. Two were already fixed on \`main\` by PRs #88 (gating knobs \`max_foveal_docs\` / \`salience_floor\`) and #90 (split inbox boosts out of chat-read salience). The remaining gap: token budget was hardcoded to 32768 in \`assembleContextInnerWithOpts\` with no kernel.yaml override or per-request mechanism, and there was no path-based exclusion beyond the hardcoded \`/archive/\` substring check.

- New \`DefaultBudget = 32768\` named constant (was a literal).
- New \`Config.DefaultBudget int\` (\`default_budget\` YAML key).
- New \`Config.ExcludeGlobs []string\` (\`exclude_globs\` YAML key — substring match against slash-normalised paths, parallel to the existing \`/archive/\` filter).
- \`EffectiveBudget()\` and \`ContextExcludeGlobs()\` accessors — RWMutex-guarded, same pattern as \`ContextGating()\`.
- \`X-Cogos-Context-Budget\` request header support in BOTH \`/v1/chat/completions\` and \`/v1/messages\` (Anthropic-compat).
- 6 new tests covering fallback, config-driven budget, explicit caller precedence, glob exclusion, helper edge cases, snapshot independence.

\`make test\` passes.

## Naming concern

The config key \`exclude_globs\` and helper \`pathMatchesExcludeGlobs\` use \"globs\" terminology but actually do \`strings.Contains\` substring matching, not \`filepath.Match\` glob matching. Worth either renaming to \`exclude_substrings\` or implementing actual glob matching. Glob matching is more useful and matches the public-facing name.

## Out of scope (filed as follow-ups)

- Per-client policies (dashboard vs Claude Code vs MCP — needs identity-routing layer).
- \`/v1/context/last\` debug endpoint.
- Hot-tunable \`exclude_globs\` via PATCH (extending the existing \`/v1/settings/context\` surface).

## Notes for review

- The hardcoded \`/archive/\` substring check still exists alongside the new mechanism. Once \`exclude_globs\` defaults include \`/archive/\`, the hardcoded check can be removed — minor maintenance debt.
- Invalid \`X-Cogos-Context-Budget\` header values are silently ignored (\`strconv.Atoi\` failure or value \`<= 0\`). Should log a warning so misconfigured clients can debug.

## Test plan

- [x] \`make test\` passes
- [x] 6 new test cases cover the new code paths
- [ ] Decide on glob vs substring semantics before non-draft (rename or implement)

### DIFF
--- a/internal/engine/config.go
+++ b/internal/engine/config.go
@@ -83,13 +83,14 @@ type Config struct {
 	// Set via default_budget in kernel.yaml.
 	DefaultBudget int
 
-	// ExcludeGlobs is a list of path substrings. Any CogDoc whose slash-normalised
-	// path contains one of these substrings is excluded from the foveated context
-	// window for chat requests. Useful to keep large or sensitive path trees
-	// (e.g. /inbox/, /archive/, /vendor/) out of ambient context without
-	// removing the files from the corpus entirely.
-	// Configured via exclude_globs in kernel.yaml.
-	ExcludeGlobs []string
+	// ExcludeSubstrings is a list of path substrings. Any CogDoc whose
+	// slash-normalised path contains one of these substrings is excluded from
+	// the foveated context window for chat requests. Useful to keep large or
+	// sensitive path trees (e.g. /inbox/, /archive/, /vendor/) out of ambient
+	// context without removing the files from the corpus entirely.
+	// Configured via exclude_substrings in kernel.yaml. Substring (not glob)
+	// semantics — implementation uses strings.Contains, not filepath.Match.
+	ExcludeSubstrings []string
 
 	// gatingMu guards the gating knobs above for hot-update via the
 	// /v1/settings/context endpoints.
@@ -151,7 +152,7 @@ type kernelConfigSection struct {
 	MaxFovealDocs         int               `yaml:"max_foveal_docs"`
 	SalienceFloor         *float64          `yaml:"salience_floor"`
 	DefaultBudget         int               `yaml:"default_budget"`
-	ExcludeGlobs          []string          `yaml:"exclude_globs"`
+	ExcludeSubstrings     []string          `yaml:"exclude_substrings"`
 	TRMWeightsPath        string            `yaml:"trm_weights_path"`
 	TRMEmbeddingsPath     string            `yaml:"trm_embeddings_path"`
 	TRMChunksPath         string            `yaml:"trm_chunks_path"`
@@ -259,8 +260,8 @@ func applyKernelSection(cfg *Config, s kernelConfigSection) {
 	if s.DefaultBudget != 0 {
 		cfg.DefaultBudget = s.DefaultBudget
 	}
-	if len(s.ExcludeGlobs) > 0 {
-		cfg.ExcludeGlobs = s.ExcludeGlobs
+	if len(s.ExcludeSubstrings) > 0 {
+		cfg.ExcludeSubstrings = s.ExcludeSubstrings
 	}
 	if s.TRMWeightsPath != "" {
 		cfg.TRMWeightsPath = s.TRMWeightsPath
@@ -329,16 +330,17 @@ func (c *Config) EffectiveBudget() int {
 	return DefaultBudget
 }
 
-// ContextExcludeGlobs returns a snapshot of the configured exclude-glob list.
-// The returned slice is safe for the caller to iterate without holding a lock.
-func (c *Config) ContextExcludeGlobs() []string {
+// ContextExcludeSubstrings returns a snapshot of the configured
+// exclude-substring list. The returned slice is safe for the caller to iterate
+// without holding a lock.
+func (c *Config) ContextExcludeSubstrings() []string {
 	c.gatingMu.RLock()
 	defer c.gatingMu.RUnlock()
-	if len(c.ExcludeGlobs) == 0 {
+	if len(c.ExcludeSubstrings) == 0 {
 		return nil
 	}
-	out := make([]string, len(c.ExcludeGlobs))
-	copy(out, c.ExcludeGlobs)
+	out := make([]string, len(c.ExcludeSubstrings))
+	copy(out, c.ExcludeSubstrings)
 	return out
 }
 
--- a/internal/engine/config.go
+++ b/internal/engine/config.go
@@ -19,6 +19,11 @@ import (
 const (
 	DefaultMaxFovealDocs = 10
 	DefaultSalienceFloor = 0.3
+	// DefaultBudget is the fallback token budget for context assembly when no
+	// per-request override (X-Cogos-Context-Budget header or MCP budget field)
+	// or kernel.yaml default_budget is provided. Matches the default provider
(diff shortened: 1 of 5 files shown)


chazmaniandinkle:
## Code review

**Recommendation: needs-changes** before un-drafting (none are large; can land same-day).

Targeted, well-scoped fix for the budget/exclusion gap of #77 left after #88/#90. Adds `default_budget` and `exclude_globs` config keys with RWMutex-guarded accessors mirroring `ContextGating()`, threads `EffectiveBudget()` into the assembler's zero-budget fallback, and accepts `X-Cogos-Context-Budget` on both OpenAI- and Anthropic-compat endpoints. +242/-4 with 6 focused tests; cleanly written.

### Strengths
- Convention-faithful: accessor pattern, locking, and apply-only-if-set YAML semantics match the existing `SalienceFloor` / `MaxFovealDocs` precedent in `applyKernelSection`.
- `ContextExcludeGlobs()` returns a defensive copy and is explicitly tested for snapshot independence.
- Precedence (caller > config > constant) is correct and pinned by `TestExplicitBudgetOverridesDefault`.
- Author flagged the glob-vs-substring naming gap themselves.

### Issues
1. **Naming contract violation (correctness).** `exclude_globs` performing `strings.Contains` is a documentation lie that will bite the first user who writes `*.md` or `**/inbox/**`. Recommendation: implement `path.Match` (or `doublestar`) before merging — the YAML key is the public surface and renaming later is a breaking change. Substring-only is fine internally; the externally-visible name should not promise glob.
2. **MCP integration gap.** `mcp_server.go:903` still hardcodes `budget = 50000` for `cog_assemble_context`. #77's acceptance criterion ("agent calling `cog_assemble_context` can pass its own budget and have it respected") is partly met (`input.Budget` is honored), but the *fallback* should also flow through `EffectiveBudget()` — or the divergence should be justified in a comment. Currently a `kernel.yaml` `default_budget: 8192` silently doesn't apply to MCP.
3. **Silent header rejection.** Malformed `X-Cogos-Context-Budget` is dropped without trace. A `slog.Warn` with the offending value (truncated) would cost nothing. Also consider an upper bound — an attacker-supplied `999999999` is honored verbatim and only clipped downstream by `evictForBudget`. Cap at e.g. `2 * EffectiveBudget()` or the provider context window.
4. **Hardcoded `/archive/` left in place.** Either remove now and seed `ExcludeGlobs` defaults with `/archive/` (and `/inbox/` if appropriate), or open a follow-up and reference it in a TODO.
5. **Test gap.** No end-to-end test that the HTTP header reaches `pkg.Budget`. A short `httptest.Server` exercise of `/v1/chat/completions` with the header set would close the loop.

### Backward compat
Clean — zero values preserve prior 32768 behavior. No migration concern.

PR #74, opened 2026-04-29
[WIP] refactor(ollama): provider rewrite — endpoint resolution + capability discovery (does not build, incomplete)

chazmaniandinkle:
## ⚠ Draft — does NOT build against current main

This PR restores the Ollama provider rewrite that was sitting in stash@{0} on the user's working tree (originally on \`feat/session-activity-channel\`, labeled "DO NOT TOUCH"). It's pushed for review and future completion, **not for merge**. CI will fail.

## What's in this PR

3 files (+311/-194):

- \`internal/engine/provider_ollama.go\` — major refactor. Uses \`ResolveLocalLLMEndpoint\` for endpoint URL resolution with sane defaults. Adds a \`listModels\` helper for capability discovery (replaces the inline \`/api/tags\` request in \`Available()\`). Cleaner separation of concerns.
- \`internal/engine/provider_openai.go\` — companion changes that share the \`ResolveLocalLLMEndpoint\` + \`listModels\` plumbing.
- \`internal/engine/tool_loop.go\` — adjacent tool-loop adjustments to support the new provider shapes.

## Why it doesn't build

The stash contains only 4 files. The Ollama rewrite depends on companion definitions that lived elsewhere on the source branch but weren't captured in the stash:

- \`ResolveLocalLLMEndpoint\` (helper in \`internal/engine\` — endpoint URL resolution with defaults)
- \`localLLMDefaultEndpoint\` (constant)
- \`KernelToolRegistry.Scoped()\` method (used by \`local_agent_harness.go\` on the source branch — a scope-filter API that pairs with the named-scope work merged in PR #68)
- The new \`OllamaProvider\` / \`OpenAICompatProvider\` structs are missing a \`Model()\` method declaration, so they no longer satisfy the \`Provider\` interface.

## To complete this PR

A real merge-ready version needs to:
1. Define \`ResolveLocalLLMEndpoint\` + \`localLLMDefaultEndpoint\` (small helpers, ~20-40 lines).
2. Add \`Scoped()\` to \`KernelToolRegistry\` (or rework the harness call sites to use a different existing API). Pairs with PR #68's named-scope substrate.
3. Re-add \`Model()\` to the new provider structs (likely just moved/dropped during the refactor).
4. Verify the rewrite still works end-to-end against an actual Ollama instance.

## Companion PR

[...]

Issue #76, opened 2026-04-29
Track: complete the Ollama provider refactor (PR #74 draft)


chazmaniandinkle:
## Summary

[PR #74](https://github.com/cogos-dev/cogos/pull/74) is a DRAFT containing a substantial Ollama provider rewrite that was preserved from a working-tree stash. The diff itself is well-intentioned (endpoint resolution helper, capability discovery via \`listModels\`, cleaner separation of concerns) but **does not build against current main** because the original source branch had companion definitions that didn't make it into the saved diff.

## What's in the draft

\`internal/engine/provider_ollama.go\`, \`internal/engine/provider_openai.go\`, \`internal/engine/tool_loop.go\` — 311 insertions, 194 deletions across 3 files.

## What's missing

To make the draft build and ship, four pieces of companion work need to land — either in this branch or as a prerequisite:

1. **Define \`ResolveLocalLLMEndpoint()\`** — helper for endpoint URL resolution with sane defaults. Probably a 20–40-line function in a new file like \`internal/engine/local_llm_endpoint.go\` or appended to an existing local-LLM file.
2. **Define \`localLLMDefaultEndpoint\`** constant — the default Ollama/LMS URL (likely \`http://localhost:11434\` and \`http://localhost:1234/v1\` shape).
3. **Add \`Scoped()\` method to \`KernelToolRegistry\`** — referenced in 3 places in \`local_agent_harness.go\` on the original branch. Pairs conceptually with the named-scope substrate landed via PR #68 (which added \`Scope\` field on \`DispatchRequest\`). One reading: \`Scoped()\` returns a sub-registry filtered to a named scope.
4. **Restore \`Model()\` method** on the new \`OllamaProvider\` and \`OpenAICompatProvider\` structs — the rewrite must have dropped or relocated the declaration that \`Provider\` interface compliance requires.

## Acceptance criteria

- \`go build ./...\` passes against current main.
- \`go test ./internal/engine/...\` passes.
- Smoke test: kernel daemon starts, dispatches successfully through both Ollama and OpenAI-compat paths to a real local model server.
- The draft PR converts to ready-for-review.

## Cost estimate

~2–3 hours of focused work. The hardest part is item 3 (\`KernelToolRegistry.Scoped()\`) since it touches the harness call sites — needs to integrate cleanly with the named-scope work from PR #68.

## Alternative: scrap and redo

[...]

PR #114, opened 2026-05-01
refactor(provider/ollama): extract listModels helper from Available

chazmaniandinkle:
## What

Extracts a `listModels(ctx context.Context) ([]string, error)` method from the inline HTTP + JSON decode logic that was embedded in `OllamaProvider.Available()`.

`Available()` previously owned the full `/api/tags` request lifecycle: build request, do HTTP, check status, decode JSON, iterate models. That logic is now a standalone helper. `Available()` calls `listModels` and checks membership — identical semantics, no inlined plumbing.

Empty model names (which Ollama can return in edge cases) are filtered in `listModels` so callers never see them.

## Why

The inline decode in `Available()` was a maintenance liability: any caller that needs the model list had to duplicate the HTTP + decode cycle. Extracting `listModels` makes it reusable (the remaining cherry-picks from #74 rely on it) and keeps `Available()` at a readable level of abstraction.

## Test

Adds `TestOllamaListModels` using `httptest.NewServer` with a fake `/api/tags` response that includes one empty-name entry. Asserts:
- two names are returned (empty filtered out)
- correct ordering preserved

All pre-existing `internal/engine` tests pass.

## Context

This is **one of seven focused cherry-picks** from closed PR #74 (`cogos-dev/cogos#74`), which attempted a sweeping Ollama refactor that did not build. The umbrella for this salvage work is issue #76. Each cherry-pick is a standalone, compilable, tested change.

### DIFF
--- a/internal/engine/provider_ollama.go
+++ b/internal/engine/provider_ollama.go
@@ -89,29 +89,13 @@ func (p *OllamaProvider) Model() string { return p.model }
 
 // Available checks if Ollama is running and the configured model is loaded.
 func (p *OllamaProvider) Available(ctx context.Context) bool {
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, p.endpoint+"/api/tags", nil)
+	models, err := p.listModels(ctx)
 	if err != nil {
-		return false
-	}
-	resp, err := p.client.Do(req)
-	if err != nil {
-		return false
-	}
-	defer resp.Body.Close()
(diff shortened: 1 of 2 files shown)


chazmaniandinkle:

Pure refactor; `Available` semantics preserved. Adds empty-name filtering and a status-code error wrap previously absent. Reusable helper for downstream PRs that may consume the model list.

No ordering dependency on the other PRs in this stack. Land first if landing sequentially.

PR #116, opened 2026-05-01
feat(tool_loop): normalize blank tool-call IDs from any provider

chazmaniandinkle:
## Summary

- Adds `normalizeToolCallIDs(providerName, iteration, calls)` to `internal/engine/tool_loop.go`, called at the top of each `RunToolLoopWithTranscript` iteration before any dispatch happens.
- Adds `sanitizeToolCallIDPrefix` helper that collapses the provider name to a safe lowercase alphanumeric-plus-dash string (falls back to `"provider"` for empty names).
- Generated ID format: `<sanitized-provider>-call-<iteration>-<index>` (e.g. `ollama-call-1-0`).
- Adds `TestNormalizeToolCallIDs` covering: all-blank input gets unique IDs, mixed blank-and-set preserves existing IDs, all-set passes through unchanged, nil input returns nil, empty provider name uses fallback prefix.

## Why provider-agnostic (vs the Ollama-specific cherry-pick)

The Ollama-specific cherry-pick (`ollamaToolCallID`) addresses ID generation inside the Ollama wire adapter. This PR is the orthogonal, tool-loop-layer fix: **any** provider (OpenAI-compat, Codex, ClaudeCode, future providers) may return tool calls with blank IDs. Normalizing at the tool-loop boundary ensures downstream dispatch code — kernel tool execution, transcript recording, `ProviderMessage.ToolCallID` correlation — never has to branch on blank IDs regardless of which provider is active.

Closes #76.

## What was NOT touched

- `KernelToolRegistry.Scoped()` is preserved (PR #74's removal of it was the semantic conflict that closed the original PR).
- `cog_read_file` and `cog_grep_files` audit-scope tools are preserved.
- No other files modified.

## Test plan

- [x] `go test ./internal/engine/ -run TestNormalizeToolCallIDs` — all 5 subtests pass
- [x] `make test` — `internal/engine` package green; root-package failure is pre-existing and unrelated to this change

PR #117, opened 2026-05-01
fix(provider/ollama): normalize tool argument JSON across object and string shapes

### LINKED ISSUE #76: Track: complete the Ollama provider refactor (PR #74 draft)

chazmaniandinkle:
## Problem

Ollama's `/api/chat` endpoint returns tool-call `arguments` in two shapes depending on the model:

- **Object shape** (standard): `{"query": "x", "limit": 5}`
- **String shape** (llama3.1, some qwen2.5 builds): `"{\"query\": \"x\", \"limit\": 5}"` — a JSON string wrapping a JSON object

The previous code either passed the raw `json.RawMessage` through unchanged or did a `json.Valid` guard that accepted the string shape as-is. Downstream callers received a double-encoded string in `ToolCall.Arguments`, which broke argument parsing in the tool loop.

Closes #76.

## Solution

- **`decodeOllamaToolArguments(raw string) interface{}`** — normalizes both shapes. Object → `map[string]any`. String-wrapped object → unwraps one layer and decodes the inner JSON. Empty → empty map. Malformed → original string (safe fallback).
- **`encodeOllamaToolArguments(v interface{}) string`** — marshals the normalized value back to a canonical JSON string. Nil → `{}`.

Both helpers are wired into:
1. **`Complete`** — when converting Ollama's response tool calls to `ToolCall` structs
2. **`buildOllamaRequest`** — when serializing stored `ToolCall.Arguments` back to Ollama for multi-turn conversations

## Test coverage

`TestDecodeOllamaToolArguments` (table-driven, parallel):
- Object input
- String input wrapping a JSON object
- Empty string
- Whitespace-only
- `null` literal
- Malformed JSON → error path

`TestEncodeOllamaToolArgumentsNil` — nil produces `{}`
`TestDecodeEncodeRoundTrip` — object path and string-wrapped path produce identical output

## Which models trigger the string-arg case

Based on community reports and the closed PR #74 discussion: **llama3.1**, **llama3.2**, and certain **qwen2.5** checkpoint builds are known to emit the string-wrapped form. The issue is model-specific and non-deterministic (same model can return either shape across calls), so normalization at the provider boundary is the correct fix.

## Diff stat

```
internal/engine/provider_ollama.go      | +62 / -8
internal/engine/provider_ollama_test.go | +128 / 0
```

### DIFF
--- a/internal/engine/provider_ollama.go
+++ b/internal/engine/provider_ollama.go
@@ -229,10 +229,11 @@ func buildOllamaRequest(model string, req *CompletionRequest, stream bool, conte
 		}
 		if m.Role == "assistant" && len(m.ToolCalls) > 0 {
 			for _, tc := range m.ToolCalls {
-				rawArgs := json.RawMessage(tc.Arguments)
-				if len(rawArgs) == 0 || !json.Valid(rawArgs) {
-					rawArgs = json.RawMessage("{}")
-				}
+				// Normalize arguments to a stable JSON object before sending
+				// back to Ollama. Some models return arguments as a JSON string
+				// rather than an object; decodeOllamaToolArguments unwraps both.
+				normalizedArgs := decodeOllamaToolArguments(tc.Arguments)
+				rawArgs := json.RawMessage(encodeOllamaToolArguments(normalizedArgs))
 				msg.ToolCalls = append(msg.ToolCalls, ollamaToolCall{
 					ID:   tc.ID,
 					Type: "function",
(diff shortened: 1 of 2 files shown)


chazmaniandinkle:

Decode-then-reencode through `decodeOllamaToolArguments` / `encodeOllamaToolArguments` correctly handles the llama3.x / qwen2.5 string-wrapped shape on both inbound (`Complete`) and outbound (`buildOllamaRequest`) paths.

### Suggestions
- The `null` literal test case asserts encoded output is `"null"` *or* `"{}"` — slightly loose; consider tightening to `{}` for consistency with the empty-input contract.
- Silent fallback to the raw string on malformed JSON could hide upstream bugs — a `slog.Debug` would help.

PR #118, opened 2026-05-01
feat(provider/ollama): map done_reason to OpenAI-style finish_reason

chazmaniandinkle:
## Summary

- Ollama's `/api/chat` response includes a `done_reason` field (`"stop"`, `"length"`, `"load"`, `"unload"`) that was silently ignored on current main — the provider always returned `StopReason: "end_turn"` regardless of what the server signalled.
- Adds `DoneReason string \`json:"done_reason,omitempty"\`` to `ollamaChatResponse` and a `mapOllamaDoneReason()` function that translates to the provider-agnostic stop-reason vocabulary used everywhere else in this codebase.
- Wires into both the streaming (`Stream`) and non-streaming (`Complete`) paths.

Closes #76.

## Mapping

| Ollama `done_reason` | `StopReason` (internal) |
|---|---|
| `stop` / `""` | `end_turn` |
| `length` | `max_tokens` |
| `tool_calls` | `tool_use` |
| `load` / `unload` | `end_turn` (model lifecycle, not generation) |
| any unknown | `end_turn` (safe default) |

## Test plan

- `TestMapOllamaDoneReason` — table-driven unit test covering all mapped values + unknown default.
- `TestOllamaCompleteMapsDoneReason` — httptest mock with `done_reason:"length"`, asserts `StopReason:"max_tokens"` in `CompletionResponse`.
- `TestOllamaStreamMapsDoneReason` — httptest mock with streaming final chunk `done_reason:"length"`, asserts `StopReason:"max_tokens"` on the final `StreamChunk`.
- `go test ./internal/engine/...` passes clean (127 LOC added, 0 deleted).

PR #121, opened 2026-05-01
fix(provider/ollama): align streaming finish_reason vocabulary to tool_use

### LINKED ISSUE #124: provider/ollama: reconcile finish_reason vocabulary between Complete and streaming paths
## Divergence

`internal/engine/provider_ollama.go` emits two different `StopReason` strings for the same logical event (the model produced tool calls), depending on which path the caller took:

- **Non-streaming `Complete`** (line 351): sets `out.StopReason = "tool_use"` — Anthropic vocabulary.
- **Streaming `Stream`** (line ~447, post-#121): sets `sc.StopReason = "tool_calls"` — OpenAI vocabulary, gated on the `sawToolCalls` tracker introduced in #121.

Both paths consume the same upstream Ollama response shape (`Message.ToolCalls`) and represent the same finish condition, but they hand callers different strings.

## Why it matters

Callers that branch on `StopReason` (the tool-loop, the OpenAI-compatible HTTP surface, downstream consumers) get inconsistent values for an equivalent tool-call response, depending solely on whether the request went through the streaming or non-streaming path. `mapStopReasonToOpenAI` in `provider.go` papers over the non-streaming case by translating `"tool_use"` -> `"tool_calls"`, but the raw `StopReason` field is observable, and the asymmetry is a footgun for any code reading it before normalization.

## Suggested resolution

Pick one vocabulary and converge both paths on it. Recommend `"tool_calls"` (OpenAI ecosystem alignment — that's where most tool-using consumer code lives, and it matches what `mapStopReasonToOpenAI` already produces).

Add a unit test asserting that `Complete` and `Stream` produce the same `StopReason` for an equivalent tool-call response from a fixture Ollama server.

---

Surfaced during review of #121.
Tracking: #76.

chazmaniandinkle:
Closes #124.

## Summary

The Ollama streaming path was emitting `StopReason: "tool_calls"` while the non-streaming `Complete()` path used `"tool_use"`. This PR aligns streaming with the codebase-wide internal vocabulary (Anthropic-style `tool_use`).

The original `sawToolCalls` tracker landed in #118 (which had a misleading title and was supposed to add `mapOllamaDoneReason()`, but actually shipped this branch's diff instead). After that merge, the streaming path's finish_reason stayed at `tool_calls`, divergent from `Complete()`. This PR completes the reconciliation.

## Changes

- `internal/engine/provider_ollama.go` — line 528: `"tool_calls"` → `"tool_use"`
- `internal/engine/provider_ollama_test.go` — `TestStreamSetsToolCallsFinishReason` and `TestStreamNoToolCallsFinishReason` asserts updated to the new vocabulary; comments rewritten to match.

## Test plan

- [x] `go test ./internal/engine/ -run TestStream` passes
- [x] `go test ./internal/engine/` passes (full engine suite green locally before push)

## Out of scope

- Adding the actual `mapOllamaDoneReason()` function that #118's body promised — `done_reason` mapping for `length`/`load`/`unload` etc. is not currently wired. Worth a follow-up issue if anyone hits the gap (today the only divergent vocabulary is the `tool_use`/`tool_calls` one this PR closes).

### DIFF
--- a/internal/engine/provider_ollama.go
+++ b/internal/engine/provider_ollama.go
@@ -525,7 +525,7 @@ func (p *OllamaProvider) Stream(ctx context.Context, req *CompletionRequest) (<-
 					Model:    model,
 				}
 				if sawToolCalls {
-					sc.StopReason = "tool_calls"
+					sc.StopReason = "tool_use"
 				}
 			}
 			select {
--- a/internal/engine/provider_ollama_test.go
+++ b/internal/engine/provider_ollama_test.go
@@ -752,9 +752,8 @@ func TestBuildOllamaRequestToolsAndToolReplies(t *testing.T) {
 }
 
 // TestStreamSetsToolCallsFinishReason asserts that when the Ollama stream
-// contains tool_call chunks, the final StreamChunk has StopReason="tool_calls".
-// Regression: issue #76 — streaming path omitted finish_reason entirely when
-// tool calls were present, breaking OpenAI-compatible consumers.
+// contains tool_call chunks, the final StreamChunk has StopReason="tool_use",
+// matching the non-streaming Complete path's vocabulary.
 func TestStreamSetsToolCallsFinishReason(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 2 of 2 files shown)


chazmaniandinkle:

Clean tracker; tests cover positive and negative cases.

### Vocabulary nit (worth a follow-up issue, not blocking)
This PR emits `"tool_calls"` (OpenAI vocabulary) while `Complete` sets `StopReason="tool_use"` (Anthropic vocabulary) at `provider_ollama.go:351`. Same provider, two vocabularies. The streaming path here is consistent with the existing OpenAI-compat streaming, so this PR is internally coherent — but worth resolving the `Complete`-vs-streaming divergence inside the Ollama provider as a separate cleanup.

PR #120, opened 2026-05-01
fix(provider/ollama): suppress tool schemas when tool_choice is "none"

chazmaniandinkle:
## Summary

- Closes #76
- When `tool_choice == "none"`, `buildOllamaRequest` was still sending the full `tools` array to Ollama. This wastes tokens on the prompt and leaves the model in an ambiguous state where it can see tool schemas but is supposed to ignore them — in practice, Ollama models will sometimes call tools anyway.
- Fix: guard the tools-building block with `req.ToolChoice != "none"`. When suppressed, the field is `nil` and omits from the JSON payload via `omitempty`. All other values (`"auto"`, `"required"`, specific tool name, or unset) are unaffected.

## What changed

- `internal/engine/provider_ollama.go`: one-line guard added to `buildOllamaRequest`
- `internal/engine/provider_ollama_test.go`: `TestBuildOllamaRequestSuppressesToolsWhenNone` added; verifies the marshaled wire body omits the `tools` key for `"none"` and includes it for `"auto"`, `"required"`, and a specific tool name

## Test plan

- [x] `go test ./internal/engine/... -run TestBuildOllamaRequest` — all subtests pass
- [x] `go test ./internal/engine/...` — full package green
- [x] Unrelated `TestQueryConstellationWithIris_FallbackToStandard` flakiness confirmed pre-existing on main (unrelated to this change)

Cherry-picked from the request-building change in PR #74; adapted cleanly to current main (no `KeepAlive` field conflict).

PR #128, opened 2026-05-02
fix(provider/openai-compat): suppress tool schemas when tool_choice is "none"

### LINKED ISSUE #127: provider/openai-compat: suppress tool schemas when tool_choice="none" (parity with provider/ollama #120)
## Description

PR #120 added a one-line guard in `internal/engine/provider_ollama.go` that suppresses the `tools` array on the wire when `tool_choice == "none"`. The OpenAI-compat provider at `internal/engine/provider_openai.go` (lines 346–364 of `buildOpenAIRequest`) currently maps `req.Tools` into `or.Tools` unconditionally and then maps `req.ToolChoice` independently, so a request with `ToolChoice == "none"` still ships the full tool schemas to the endpoint.

## Why it matters

When an agent explicitly sets `tool_choice="none"`, sending tool schemas in the request is wasteful (extra prompt tokens on every turn) and counterproductive: many models bias toward tool use when schemas are present in context, even when told not to call them. The Ollama provider already pays this cost; the OpenAI-compat path should match.

## Acceptance criteria

- In `buildOpenAIRequest`, gate the tools-mapping block on `req.ToolChoice != "none"` (mirror the shape of the Ollama guard introduced in #120: `if len(req.Tools) > 0 && req.ToolChoice != "none"`).
- Add a test in `provider_openai_test.go` mirroring the patterns in `provider_ollama_test.go` — specifically `TestBuildOllamaRequestSuppressesToolsWhenNone` — covering: (a) `tool_choice="none"` omits the `tools` key from the JSON wire body; (b) `auto`, `required`, and a specific tool name still include tools; (c) no tools defined → field absent regardless of `tool_choice`.
- No behavior change for any value of `tool_choice` other than `"none"`.

## Reference

PR #120 review (`cogos-dev/cogos#120`).

chazmaniandinkle:
## Summary

- Gates the `tools` array in `buildOpenAIRequest` behind `req.ToolChoice != "none"`, matching the guard introduced for the Ollama provider in #120.
- Adds `TestBuildOpenAIRequestSuppressesToolsWhenNone` with five subtests mirroring `TestBuildOllamaRequestSuppressesToolsWhenNone` from the Ollama provider.

## Changes

- `internal/engine/provider_openai.go`: changed `if len(req.Tools) > 0` to `if len(req.Tools) > 0 && req.ToolChoice != "none"` in `buildOpenAIRequest`.
- `internal/engine/provider_openai_test.go`: added `TestBuildOpenAIRequestSuppressesToolsWhenNone` covering: `none` omits the `tools` key from the JSON wire body; `auto`, `required`, and a specific tool name still include tools; no tools defined omits the field regardless of `tool_choice`.

## Test plan

- [ ] `go test ./internal/engine/ -count=1 -timeout 90s` passes (verified locally).
- [ ] `TestBuildOpenAIRequestSuppressesToolsWhenNone` and all 5 subtests pass.
- [ ] Existing `TestBuildOpenAIRequestTools` (auto path) continues to pass — no regression.

## Out of scope

- Streaming path: `buildOpenAIRequest` is shared by both `Complete` and `Stream`, so the guard covers both without any additional changes.
- No changes to the `tool_choice` pass-through logic — `"none"` still flows to the wire correctly when no tools are present or when tools are suppressed.

Closes #127

### DIFF
--- a/internal/engine/provider_openai.go
+++ b/internal/engine/provider_openai.go
@@ -343,8 +343,10 @@ func buildOpenAIRequest(model string, req *CompletionRequest, stream bool, maxTo
 		or.MaxTokens = req.MaxTokens
 	}
 
-	// Map tools.
-	if len(req.Tools) > 0 {
+	// Map tools. Skip when tool_choice is "none" — the caller has explicitly
+	// opted out of tool use for this turn; sending schemas wastes tokens and
+	// can bias models toward tool calls they should ignore.
+	if len(req.Tools) > 0 && req.ToolChoice != "none" {
 		or.Tools = make([]openaiTool, len(req.Tools))
 		for i, t := range req.Tools {
 			or.Tools[i] = openaiTool{
--- a/internal/engine/provider_openai_test.go
+++ b/internal/engine/provider_openai_test.go
@@ -168,6 +168,66 @@ func TestBuildOpenAIRequestTools(t *testing.T) {
 	}
 }
 
+// TestBuildOpenAIRequestSuppressesToolsWhenNone verifies that the tools array
+// is omitted from the wire body when tool_choice is "none", and that it is
+// included for other values ("auto", "required", specific tool name).
+func TestBuildOpenAIRequestSuppressesToolsWhenNone(t *testing.T) {
+	t.Parallel()
+
+	tools := []ToolDefinition{
+		{
+			Name:        "search",
+			Description: "Search the memory index",
+			InputSchema: map[string]interface{}{"type": "object"},
+		},
+	}
+	msgs := []ProviderMessage{{Role: "user", Content: "hi"}}
+
+	// tool_choice == "none" → tools field must be absent from the wire body.
+	t.Run("none suppresses tools", func(t *testing.T) {
+		t.Parallel()
+		req := &CompletionRequest{Messages: msgs, Tools: tools, ToolChoice: "none"}
+		r := buildOpenAIRequest("m", req, false, 4096)
+		if len(r.Tools) != 0 {
+			t.Errorf("tools len = %d; want 0 when tool_choice is none", len(r.Tools))
+		}
+		body, err := json.Marshal(r)
+		if err != nil {
+			t.Fatalf("marshal: %v", err)
+		}
+		var raw map[string]json.RawMessage
+		if err := json.Unmarshal(body, &raw); err != nil {
+			t.Fatalf("unmarshal: %v", err)
+		}
+		if _, present := raw["tools"]; present {
+			t.Errorf("wire body contains 'tools' key; want omitted when tool_choice is none")
+		}
+	})
+
+	// Other tool_choice values must still send the tools schema.
+	for _, tc := range []string{"auto", "required", "search"} {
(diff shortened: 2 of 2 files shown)


chazmaniandinkle (commented):
## Verdict

minor-follow-ups

## Acceptance

All three acceptance criteria from #127 are met:

- **Guard shape**: `provider_openai.go:349` uses exactly `if len(req.Tools) > 0 && req.ToolChoice != "none"`, matching the Ollama guard at `provider_ollama.go:291`.
- **Test coverage**: `TestBuildOpenAIRequestSuppressesToolsWhenNone` covers all five cases the issue specified: `none` suppresses tools (struct-level and wire-level), `auto`/`required`/named tool preserve tools, and no-tools-defined omits the field regardless of `tool_choice`.
- **No behavior change for other values**: the `ToolChoice` pass-through block at `provider_openai.go:364-365` is untouched — `"none"` still appears on the wire as `tool_choice: "none"` when tools are suppressed.

## Code

**Correctness**: The guard is correct and the asymmetry with Ollama is handled properly. Ollama has no `tool_choice` wire field; the OpenAI-compat struct does (`provider_openai.go:202`), so `"none"` correctly passes through on the wire body even when `or.Tools` is nil. This is the right behavior per the OpenAI spec.

**Comment length** (`provider_openai.go:346-348`): The 3-line block comment explains the why adequately, and "why is non-obvious" clears the bar in CONTRIBUTING.md (`§ Code style`). However, the Ollama guard shipped with zero comment. The OpenAI comment is fine to keep, but it's three lines where one would suffice. Not blocking, just noting the inconsistency.

No other style or correctness issues. `gofmt`/`go vet` surface is clean.

## Tests

**Passes**: All 5 subtests pass locally (`go test ./internal/engine/ -run TestBuildOpenAIRequestSuppressesToolsWhenNone -count=1`).

**Parallelism**: The new test adds `t.Parallel()` inside every subtest. The reference Ollama test (`TestBuildOllamaRequestSuppressesToolsWhenNone`) does not. The OpenAI version is strictly better here; no concern.

**One gap**: The "none suppresses tools" subtest (`provider_openai_test.go:187-205`) verifies that `raw["tools"]` is absent from the wire body, but does not assert that `raw["tool_choice"]` equals `"none"`. Since the purpose of the guard is to suppress schemas while leaving `tool_choice` intact, asserting the presence and value of `tool_choice` in the suppression subtest would make the test's intent fully self-documenting. Not blocking — the existing `TestBuildOpenAIRequestTools` covers the pass-through path — but it would close the gap cleanly.

## Follow-ups

- **Ollama test parallelism** (`provider_ollama_test.go:886-924`): The Ollama suppression test's subtests lack `t.Parallel()`. Now that the OpenAI version has set the better pattern, the Ollama test is worth updating for consistency. Worth a small cleanup issue.
- **`tool_choice` wire assertion**: Adding `if v, ok := raw["tool_choice"]; !ok || string(v) != `"`none`"` ` to the existing "none suppresses tools" subtest would fully document the invariant. Low-priority, could be done in place or deferred.

PR #129, opened 2026-05-02
feat(bus): add ?since=<seq> query param to bound replay on /v1/bus/{id}/stream

### LINKED ISSUE #126: bus: add ?since=<seq> query parameter to /v1/bus/{id}/stream to bound replay
## Problem

`handleBusStream` in `internal/engine/serve_bus.go` (added in #111) replays
the full bus history on every connect by calling `s.busSessions.ReadEvents(busID)`
with no upper bound, then iterating every returned event and writing it to the
SSE stream (the `if !noReplay && s.busSessions != nil { … }` block — replay
loop with `fmt.Fprintf(w, "data: %s\n\n", b)` and a per-frame
`SetWriteDeadline(busSSEWriteDeadline)` reset, flushed only after the loop).

The only escape hatch today is `?no_replay=1`, which is all-or-nothing.

This has three problems for any long-lived bus:

1. **O(N) memory + I/O per reconnect.** `ReadEvents` reads the entire on-disk
   log; for buses with thousands of events every reconnect re-reads and
   re-marshals all of them.
2. **DoS vector.** An unauthenticated/cheap reconnect loop forces full-history
   replay each time.
3. **Write-deadline pressure during replay.** The handler resets the 10 s
   deadline per `Fprintf` but only `Flush()`es after the loop completes, so a
   slow consumer can stall the entire replay window without surfacing useful
   backpressure.

## Acceptance criteria

- Handler accepts `?since=N` where `N` is an integer sequence number.
- Replay yields only events with `seq > N` (in order).
- When `N` equals the current highest `seq`, replay is omitted entirely and
  the stream proceeds straight to live (`connected` + heartbeat + live events).
- When `N` is greater than the current highest `seq`, the handler returns
  `416 Range Not Satisfiable` (or similar, with a JSON error body) before
  upgrading to SSE.
- `since` and `no_replay=1` are mutually exclusive — passing both is a
  `400 Bad Request`.
- Test coverage: replay-from-mid, replay-from-tip (no replay), replay-past-tip
  (416), and the mutual-exclusion case.

## Out of scope

Cursor-based (`?after=<event_id>`) and time-based (`?since_ts=<RFC3339>`)
variants. File those separately if needed.

## Urgency

Should land before `/v1/bus/{id}/stream` is consumed by anything beyond the
`bus_watch` CLI. Dashboard / hook consumers will reconnect frequently and
amplify the unbounded-replay cost.

## Reference

- PR #111 review (the unbounded-replay concern was flagged as follow-up, not a
  merge blocker).
- Closes the open follow-up scope of #51.

chazmaniandinkle:
## Summary

- Adds `?since=N` query parameter to `GET /v1/bus/{id}/stream` so clients can reconnect and replay only events they haven't seen, rather than forcing a full-history replay on every connect.
- `?since=N` and `?no_replay=1` are mutually exclusive (400 if both supplied).
- `?since=N` where N exceeds the current tip returns 416 Range Not Satisfiable before the SSE upgrade, so the client gets a clean error rather than a silent empty stream.

## Changes

- `internal/engine/serve_bus.go` — `handleBusStream`: parse `since` query param; validate mutual exclusion with `no_replay`; 416 on over-range; filter replay slice to `seq > sinceSeq`.
- `internal/engine/serve_bus_test.go` — four new tests matching the issue's named cases:
  - `TestBusStreamSinceReplayFromMid` — since=2 on a 5-event bus replays seq 3,4,5 only
  - `TestBusStreamSinceReplayFromTip` — since=tip yields zero replay frames, live event still arrives
  - `TestBusStreamSincePastTip` — since=99 on a 2-event bus returns 416 with JSON error body
  - `TestBusStreamSinceMutualExclusion` — since=1&no_replay=1 returns 400 with JSON error body

## Test plan

- [ ] `go test ./internal/engine/ -run TestBusStreamSince -v` — all four new tests pass
- [ ] `go test ./internal/engine/ -count=1 -timeout 90s` — full engine suite passes

## Out of scope

- Cursor-based (`?after=<event_id>`) and time-based (`?since_ts=<RFC3339>`) variants (noted in the issue as separate follow-ups).
- Flush-per-frame during replay (issue #126 problem 3, backpressure); orthogonal concern, filed separately if needed.

Closes #126

### DIFF
--- a/internal/engine/serve_bus.go
+++ b/internal/engine/serve_bus.go
@@ -458,9 +458,14 @@ const (
 //
 // Replay query params:
 //
-//	no_replay=1  skip historical events (only live from this point forward)
-//	consumer_id  optional identity for dedup (replaces any prior connection
-//	             with the same ID)
+//	no_replay=1   skip historical events (only live from this point forward)
+//	since=<N>     replay only events with seq > N; mutually exclusive with no_replay=1
+//	consumer_id   optional identity for dedup (replaces any prior connection
+//	              with the same ID)
+//
+// since=<N> edge cases:
+//   - N == current highest seq: replay is skipped, stream proceeds live
+//   - N > current highest seq: 416 Range Not Satisfiable (before SSE upgrade)
 //
 // The stream runs until the client disconnects or the server shuts down.
 // There is no max_duration cap on bus streams — they are intended for
@@ -478,13 +483,44 @@ func (s *Server) handleBusStream(w http.ResponseWriter, r *http.Request, busID s
 		return
 	}
 
+	q := r.URL.Query()
+	noReplay := q.Get("no_replay") == "1"
+	sinceRaw := q.Get("since")
+	consumerID := q.Get("consumer_id")
+
+	// since and no_replay are mutually exclusive.
+	if sinceRaw != "" && noReplay {
+		http.Error(w, `{"error":"since and no_replay are mutually exclusive"}`, http.StatusBadRequest)
+		return
+	}
+
+	var sinceSeq int
+	hasSince := sinceRaw != ""
+	if hasSince {
+		n, err := strconv.Atoi(sinceRaw)
+		if err != nil || n < 0 {
+			http.Error(w, `{"error":"since must be a non-negative integer"}`, http.StatusBadRequest)
+			return
+		}
+		sinceSeq = n
+	}
+
+	// Validate since range before upgrading to SSE — must not exceed current tip.
+	if hasSince && s.busSessions != nil {
+		events, err := s.busSessions.ReadEvents(busID)
+		if err == nil && len(events) > 0 {
+			maxSeq := events[len(events)-1].Seq
+			if sinceSeq > maxSeq {
+				http.Error(w, `{"error":"since exceeds current tip","code":"range_not_satisfiable"}`, http.StatusRequestedRangeNotSatisfiable)
+				return
+			}
+		}
+	}
+
 	w.Header().Set("Content-Type", "text/event-stream")
 	w.Header().Set("Cache-Control", "no-cache")
 	w.Header().Set("X-Accel-Buffering", "no")
 	w.Header().Set("Connection", "keep-alive")
-
-	noReplay := r.URL.Query().Get("no_replay") == "1"
-	consumerID := r.URL.Query().Get("consumer_id")
 
 	rc := http.NewResponseController(w)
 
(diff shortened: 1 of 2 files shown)


chazmaniandinkle (commented):
## Verdict

`minor-follow-ups` — logic is correct for the tested paths; two non-blocking issues worth noting before the feature sees heavy production use.

---

## Acceptance

All four named acceptance criteria from #126 are met:

| Criterion | Status |
|---|---|
| `?since=N` replays only `seq > N` | ✅ `e.Seq > sinceSeq` at line 553 |
| `since=tip` → zero replay, live stream proceeds | ✅ `TestBusStreamSinceReplayFromTip` passes |
| `since > tip` → 416 before SSE upgrade | ✅ `http.Error` at line 514, before headers at line 520 |
| `since` + `no_replay=1` → 400 | ✅ mutual-exclusion guard at line 492 |

All four named tests are present and pass (`TestBusStreamSinceReplayFromMid`, `TestBusStreamSinceReplayFromTip`, `TestBusStreamSincePastTip`, `TestBusStreamSinceMutualExclusion`).

---

## Code

**Correctness: one real gap**

`serve_bus.go:511` — the 416 guard condition is:

```go
if err == nil && len(events) > 0 {
```

When the bus exists but has zero events, `len(events) == 0` causes the entire check to be skipped. A request with `?since=99` against an empty bus silently falls through to the SSE upgrade and produces an empty replay rather than a 416. This contradicts the acceptance criterion: "N greater than the current highest seq → 416." An empty bus has no tip; any positive `since` value is therefore over-range.

The fix is to treat an empty bus as having a tip of 0:

```go
maxSeq := 0
if len(events) > 0 {
    maxSeq = events[len(events)-1].Seq
}
if sinceSeq > maxSeq {
    // 416
}
```

Note: `since=0` on an empty bus should be allowed (means "replay all" = replay nothing), so the check should be `sinceSeq > maxSeq`, not `sinceSeq >= maxSeq`. That's already what the code does — it just needs the empty-bus case wired in.

**Performance: double `ReadEvents` call**

`serve_bus.go:510` and `serve_bus.go:548` both call `s.busSessions.ReadEvents(busID)`. The first call's result is discarded. For buses with many events this matters — the pre-check scan and the replay scan read the same file twice. Not a correctness issue, but it's unnecessary work. The events slice from the pre-check could be threaded through, or the 416 check could be deferred until inside the existing replay block (with early-return before SSE write if 416 is needed — but at that point headers are already sent, so the current structure of checking before the SSE upgrade is the right shape; the fix is to reuse the slice, not to move the check).

**Style and naming**

No violations of CONTRIBUTING.md conventions. Error messages are lowercase, no trailing punctuation. The added doc comment block is informative about the why (edge cases). `hasSince` boolean parallels the existing `noReplay` pattern cleanly.

**Sequence comparison semantics**

`e.Seq > sinceSeq` is correct per the spec ("events with `seq > N`"). `since=2` correctly excludes the event at seq 2.

**Integer parsing**

`strconv.Atoi` + `n < 0` guard correctly rejects negatives and non-numeric strings. `since=0` is permitted and means "replay all events with seq > 0" (i.e., full replay), which is reasonable and consistent.

---

## Tests

Four named tests pass. Test coverage for the four issue-specified cases is complete.

**Gaps worth noting (not blocking):**

- No test for `since=N` on an **empty bus** (`len(events) == 0`). This is where the silent-pass bug described above lives.
- No test for `since=0` (should behave like full replay).
- No test for a non-numeric `since` (e.g., `since=abc` → 400). The validation is present; a test would pin it.
- `TestBusStreamSinceReplayFromMid` uses a 3 s inner timeout + `goto done` to exit the read loop. This is correct but slightly fragile if CI is slow — the test passes because seq values are checked after the loop, not because the stream ended cleanly. Fine for now.

---

## Follow-ups

1. **Fix empty-bus 416 gap** (small, isolated change in the range-check block at `serve_bus.go:511`).
2. **Eliminate double `ReadEvents`** — thread the pre-check result through to the replay block. Mechanical refactor, zero behavior change.
3. These are the only items; both can land as a single small follow-up commit or be rolled into this PR if the author prefers — neither is a merge blocker for the tested paths.

PR #130, opened 2026-05-02
feat(engine): add substrate-level env isolation for provider child processes

### LINKED ISSUE #59: Add substrate-level env isolation for provider child processes
## Problem

CogOS currently launches provider-backed CLI subprocesses (`claude`, `pi`, `codex`) with the daemon environment unless each provider manually overrides it. That lets ingress/control-plane variables leak into egress/provider workers.

Observed failure mode: `cog claude` intentionally routes the outer Claude Code client through the kernel by setting `ANTHROPIC_BASE_URL=http://localhost:6931`. If the kernel process, or a kernel-spawned provider child, inherits that variable, a provider-spawned `claude` process can route back into the kernel. In logs this can appear as paired `procmgr: tracked` entries for one visible Claude Code prompt.

This is related to, but broader than, the earlier Codex-provider availability routing issue in #57.

## Goal

Create a substrate-level process boundary rule: provider child processes should not inherit ingress routing environment by default.

`cog claude` remains an ingress adapter and should keep setting `ANTHROPIC_BASE_URL` when the kernel is up. Kernel-spawned provider workers are egress workers and should get a sanitized provider-child environment unless explicitly configured otherwise.

## Proposed design

Add a shared command factory for provider subprocesses, owned by the engine layer. Possible API shape:

```go
func NewProviderCommandContext(ctx context.Context, binary string, args ...string) *exec.Cmd
```

or, if this should sit under process management:

```go
func (pm *ProcessManager) NewCommandContext(ctx context.Context, opts ManagedCommandOpts, binary string, args ...string) *exec.Cmd
```

with options like:

```go
type ManagedCommandOpts struct {
    EnvPolicy EnvPolicy
    ExtraEnv  []string
    Dir       string
}

type EnvPolicy string

const (
    EnvPolicyProviderChild EnvPolicy = "provider_child"
    EnvPolicyInherit       EnvPolicy = "inherit"
)
```

Provider-backed CLI subprocesses should default to `EnvPolicyProviderChild`.

## Provider child env policy

Start from `os.Environ()`, then strip known ingress/proxy/control-plane variables that can redirect model clients back into CogOS. Required initial strip list:

- `ANTHROPIC_BASE_URL`
- `ANTHROPIC_AUTH_TOKEN`
- `OPENAI_BASE_URL`
- `OPENAI_API_BASE`
- `OPENAI_COMPAT_BASE_URL`

Preserve unrelated env such as `PATH`, `HOME`, keychain/session variables, and tool-specific state. API key variables such as `ANTHROPIC_API_KEY` and `OPENAI_API_KEY` should probably remain unless there is a demonstrated loop risk. Claude Code normally uses OAuth/keychain, so the required fix is specifically that kernel-spawned `claude` does not inherit `ANTHROPIC_BASE_URL`.

If `ExtraEnv` is added, append it after sanitization so explicit provider config can opt back into a value intentionally.

## Initial call sites

Update CLI-backed providers to use the shared factory instead of direct `exec.CommandContext` with default inherited env:

- `internal/engine/provider_claudecode.go`
- `internal/engine/provider_pi.go`
- `internal/engine/provider_codex.go`

Keep `cog.go` / `cmdClaude` behavior as-is because it is the ingress launcher and intentionally points Claude Code at the kernel.

## Tests

Add unit tests for the shared env policy:

1. Parent env contains `ANTHROPIC_BASE_URL=http://localhost:6931`; provider-child env omits it.
2. Parent env contains unrelated vars such as `PATH`; provider-child env preserves them.
3. If `ExtraEnv` is supported, explicit env appended after sanitization wins.
4. `EnvPolicyInherit` preserves the original env.

Add a Claude provider regression test:

- Use a fake `claude` binary that records its environment to a temp file and emits valid provider output.
- Set parent env `ANTHROPIC_BASE_URL=http://localhost:6931`.
- Invoke `ClaudeCodeProvider.Complete` or `Stream`.
- Assert the fake child did not receive `ANTHROPIC_BASE_URL`.

Optional integration coverage:

- One Anthropic `/v1/messages` request through the kernel should produce one provider subprocess for the selected provider, absent tool-loop recursion.

## Acceptance criteria

- `cog claude` still routes the outer Claude Code client through the kernel.
- Kernel-spawned Claude Code subprocesses do not inherit `ANTHROPIC_BASE_URL`.
- The duplicate/recursive provider subprocess behavior caused by env loopback is eliminated.
- CLI-backed providers use a single shared provider-child command/env path.
- Tests cover both the shared env policy and the Claude provider regression.

## Notes

The tactical patch would be to strip `ANTHROPIC_BASE_URL` directly in `ClaudeCodeProvider`, but the substrate-native fix is to make process env isolation a shared boundary invariant for provider child processes.

chazmaniandinkle:
## Summary

- Introduces `EnvPolicy`, `ManagedCommandOpts`, and `NewProviderCommandContext` as a shared command factory for all CLI-backed provider subprocesses.
- `EnvPolicyProviderChild` strips five known ingress/proxy variables (`ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, `OPENAI_BASE_URL`, `OPENAI_API_BASE`, `OPENAI_COMPAT_BASE_URL`) from `os.Environ()` before launching a provider worker, preventing kernel routing state from leaking into egress workers.
- `ExtraEnv` appends after sanitization so providers can opt back into a specific value intentionally.
- `EnvPolicyInherit` leaves `cmd.Env` nil (unchanged from current behavior) for callers that need the full parent env.

## Changes

- `internal/engine/provider_env.go` — new shared command factory: `EnvPolicy`, `ManagedCommandOpts`, `NewProviderCommandContext`, `providerChildEnv`, `isIngressVar`.
- `internal/engine/provider_env_test.go` — unit tests for all four issue-specified scenarios plus the `ClaudeCodeProvider` regression test.
- `internal/engine/provider_claudecode.go` — `Complete`, `Stream`, and `SpawnBackground` use `NewProviderCommandContext`.
- `internal/engine/provider_pi.go` — `Complete` and `Stream` use `NewProviderCommandContext`.
- `internal/engine/provider_codex.go` — `Complete` and `Stream` use `NewProviderCommandContext`.

## Test plan

- [ ] `go test ./internal/engine/ -count=1 -timeout 90s` passes (verified locally).
- [ ] `TestProviderChildEnv_StripsIngressVars` — parent env has `ANTHROPIC_BASE_URL`; child env omits it.
- [ ] `TestProviderChildEnv_PreservesUnrelatedVars` — `PATH` survives sanitization.
- [ ] `TestProviderChildEnv_ExtraEnvAppendsAfterSanitization` — ExtraEnv wins over stripped key.
- [ ] `TestEnvPolicyInherit_PreservesEnv` — `cmd.Env` is nil under `EnvPolicyInherit`.
- [ ] `TestEnvPolicyProviderChild_StripsAllIngressKeys` — all five ingress keys stripped.
- [ ] `TestClaudeCodeComplete_DoesNotInheritBaseURL` — fake `claude` binary records its env; asserts `ANTHROPIC_BASE_URL` absent.

## Out of scope

- `Ping` methods on all three providers still use raw `exec.CommandContext` — they check binary availability only and do not run model inference, so env isolation is not needed there.
- `Available` helpers use `exec.LookPath` only — no subprocess, no isolation needed.
- Multi-provider integration test (one `/v1/messages` request producing one subprocess, no tool-loop recursion) noted as optional in the issue is deferred; unit and regression coverage satisfies the required acceptance criteria.
- CodexProvider has no `ProcessManager` integration; that is a pre-existing design gap outside this issue's scope.

Closes #59

### DIFF
--- a/internal/engine/provider_claudecode.go
+++ b/internal/engine/provider_claudecode.go
@@ -155,7 +155,7 @@ func (p *ClaudeCodeProvider) Complete(ctx context.Context, req *CompletionReques
 		"--include-partial-messages",
 	)
 
-	cmd := exec.CommandContext(ctx, p.cliBinary, args...)
+	cmd := NewProviderCommandContext(ctx, ManagedCommandOpts{EnvPolicy: EnvPolicyProviderChild}, p.cliBinary, args...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	cmd.WaitDelay = claudeCodeKillGrace
 	cmd.Cancel = func() error {
@@ -260,7 +260,7 @@ func (p *ClaudeCodeProvider) Stream(ctx context.Context, req *CompletionRequest)
 		"--include-partial-messages",
 	)
 
-	cmd := exec.CommandContext(ctx, p.cliBinary, args...)
+	cmd := NewProviderCommandContext(ctx, ManagedCommandOpts{EnvPolicy: EnvPolicyProviderChild}, p.cliBinary, args...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	cmd.WaitDelay = claudeCodeKillGrace
 	cmd.Cancel = func() error {
@@ -723,7 +723,7 @@ func (p *ClaudeCodeProvider) SpawnBackground(opts BackgroundTaskOpts) (string, e
 		ctx, cancel = context.WithTimeout(context.Background(), opts.Timeout)
 	}
 
-	cmd := exec.CommandContext(ctx, p.cliBinary, args...)
+	cmd := NewProviderCommandContext(ctx, ManagedCommandOpts{EnvPolicy: EnvPolicyProviderChild, Dir: opts.WorkDir}, p.cliBinary, args...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	cmd.WaitDelay = claudeCodeKillGrace
 	cmd.Cancel = func() error {
@@ -737,9 +737,6 @@ func (p *ClaudeCodeProvider) SpawnBackground(opts BackgroundTaskOpts) (string, e
 		return syscall.Kill(-cmd.Process.Pid, syscall.SIGTERM)
 	}
 	cmd.Stdin = strings.NewReader(opts.Prompt)
-	if opts.WorkDir != "" {
-		cmd.Dir = opts.WorkDir
-	}
 
 	proc := p.procMgr.Track(cmd, ManagedProcessOpts{
 		Kind:            ProcessBackground,
--- a/internal/engine/provider_codex.go
+++ b/internal/engine/provider_codex.go
@@ -217,10 +217,7 @@ func (p *CodexProvider) Complete(ctx context.Context, req *CompletionRequest) (*
 	if err != nil {
 		return nil, fmt.Errorf("codex binary not available: %w", err)
 	}
-	cmd := exec.CommandContext(ctx, binary, args...)
-	if p.workDir != "" {
-		cmd.Dir = p.workDir
-	}
+	cmd := NewProviderCommandContext(ctx, ManagedCommandOpts{EnvPolicy: EnvPolicyProviderChild, Dir: p.workDir}, binary, args...)
 
 	out, err := cmd.Output()
 	if err != nil {
@@ -276,10 +273,7 @@ func (p *CodexProvider) Stream(ctx context.Context, req *CompletionRequest) (<-c
 	if err != nil {
 		return nil, fmt.Errorf("codex binary not available: %w", err)
 	}
-	cmd := exec.CommandContext(ctx, binary, args...)
-	if p.workDir != "" {
-		cmd.Dir = p.workDir
-	}
+	cmd := NewProviderCommandContext(ctx, ManagedCommandOpts{EnvPolicy: EnvPolicyProviderChild, Dir: p.workDir}, binary, args...)
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
--- a/internal/engine/provider_env.go
+++ b/internal/engine/provider_env.go
@@ -1,0 +1,81 @@
+package engine
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// EnvPolicy controls what environment a provider child process inherits.
+type EnvPolicy string
+
+const (
+	// EnvPolicyProviderChild starts from os.Environ() and strips known
+	// ingress/proxy variables that can redirect model clients back into CogOS.
+	EnvPolicyProviderChild EnvPolicy = "provider_child"
+
+	// EnvPolicyInherit passes the current process environment unchanged.
+	EnvPolicyInherit EnvPolicy = "inherit"
+)
+
+// ingressVars are environment keys stripped under EnvPolicyProviderChild.
+// These are routing/proxy variables that can cause a kernel-spawned provider
+// subprocess to route its own model calls back through the kernel.
+var ingressVars = []string{
+	"ANTHROPIC_BASE_URL",
+	"ANTHROPIC_AUTH_TOKEN",
+	"OPENAI_BASE_URL",
+	"OPENAI_API_BASE",
+	"OPENAI_COMPAT_BASE_URL",
+}
+
+// ManagedCommandOpts configures environment and working directory for a
+// provider child process created via NewProviderCommandContext.
+type ManagedCommandOpts struct {
+	EnvPolicy EnvPolicy
+	ExtraEnv  []string
+	Dir       string
+}
+
+// NewProviderCommandContext creates an exec.Cmd for a provider child process,
+// applying the env policy before the process starts.
+func NewProviderCommandContext(ctx context.Context, opts ManagedCommandOpts, binary string, args ...string) *exec.Cmd {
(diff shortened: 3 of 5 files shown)


chazmaniandinkle (commented):
## Verdict

minor-follow-ups

## Acceptance

Issue #59 specified:
- A shared command factory with `EnvPolicy`, `ManagedCommandOpts`, `NewProviderCommandContext` — **implemented** exactly as proposed in `provider_env.go`.
- Strip list of five ingress vars — **implemented** at `provider_env.go:25-31`.
- Call sites: `provider_claudecode.go` (Complete, Stream, SpawnBackground), `provider_pi.go` (Complete, Stream), `provider_codex.go` (Complete, Stream) — **all six paths use the factory**.
- `ExtraEnv` appended after sanitization — **implemented** and tested.
- Unit tests for all four env-policy scenarios — **implemented** (`TestProviderChildEnv_StripsIngressVars`, `TestProviderChildEnv_PreservesUnrelatedVars`, `TestProviderChildEnv_ExtraEnvAppendsAfterSanitization`, `TestEnvPolicyInherit_PreservesEnv`, `TestEnvPolicyProviderChild_StripsAllIngressKeys`).
- Claude provider regression test with a fake binary — **implemented** (`TestClaudeCodeComplete_DoesNotInheritBaseURL`). The fake binary writes its env to a temp file; the test reads that file and asserts `ANTHROPIC_BASE_URL` absent. This exercises the actual child process env, not just `cmd.Env` at the parent layer.
- `cog claude` ingress behavior unchanged — `cog.go:5522-5538` sets `ANTHROPIC_BASE_URL` explicitly and is untouched. Confirmed.

All required acceptance criteria are satisfied. The optional integration test is correctly deferred.

## Code

**Strip list completeness.** The five named keys are present. Two gaps worth noting as follow-ups (neither is a regression introduced by this PR — both were absent before):

1. `ANTHROPIC_AUTH_TOKEN` is in the list but appears unused by the Claude Code CLI in practice; keeping it is fine and matches the issue spec.
2. `HTTP_PROXY` / `HTTPS_PROXY` / `http_proxy` / `https_proxy` / `NO_PROXY` are not in the list. If a provider subprocess honors these (e.g. codex CLI), it could route through an intercepting proxy. Issue #59 does not require them, and there is no demonstrated loop risk, but the strip list comment says "routing/proxy variables" — the proxy family is at least worth a named follow-up.
3. `OPENAI_API_KEY` is preserved, which matches the explicit guidance in issue #59 ("should probably remain unless there is a demonstrated loop risk"). No issue here.

**Provider paths.** Every inference-bearing `exec.CommandContext` call in `internal/engine/provider_{claudecode,pi,codex}.go` now goes through `NewProviderCommandContext` with `EnvPolicyProviderChild`. The three remaining raw `exec.CommandContext` calls in those files are all `Ping`/`--version`/`--help` invocations that check binary availability and do not run model inference — consistent with what the PR's "Out of scope" section states.

**`NewProviderCommandContext` default behavior.** The `switch` falls through to `providerChildEnv` for any unknown `EnvPolicy` value, not just the named `EnvPolicyProviderChild`. This means a typo in a future `EnvPolicy` constant silently strips env rather than inheriting. Consider inverting: strip only on `EnvPolicyProviderChild`, inherit on everything else. Not a correctness issue today because only two constants exist.

**`harness/` module.** `harness/harness.go:406,853` and `harness/codex.go:340,487` spawn `claude` and `codex` subprocesses using `filteredEnviron()`, which strips only `CLAUDECODE=` — it does not strip `ANTHROPIC_BASE_URL`. The harness is a separate Go module (`github.com/cogos-dev/cogos/harness`) imported by the kernel via a `replace` directive. These paths are outside this PR's stated scope and outside issue #59's named call sites. However, the same env-loopback bug can trigger through the harness path when the kernel routes an inference request there. This is not a regression introduced by this PR, but the fix is incomplete while that path exists.

## Tests

All six new tests pass (`go test ./internal/engine/ -run "Env|Provider|Claude" -count=1`, confirmed locally). The regression test (`TestClaudeCodeComplete_DoesNotInheritBaseURL`) genuinely exercises the child process: the fake `claude` script calls `env > $ENVFILE`, and the test reads `$ENVFILE` after `Complete` returns and asserts `ANTHROPIC_BASE_URL` absent. This is real end-to-end verification of the env strip, not just a `cmd.Env` field assertion. Coverage is adequate for the scoped work.

No test covers the `harness` module's `filteredEnviron` path, but that is pre-existing and out of this PR's scope.

## Follow-ups

1. **`harness/filteredEnviron` does not strip `ANTHROPIC_BASE_URL`** — `harness/harness.go:78-91`, `harness/codex.go:340,487`. The harness is a separate module and was not named in issue #59, but it exposes the same class of env-loopback risk. Recommend a follow-up issue to either adopt `NewProviderCommandContext` or add the ingress strip list to `filteredEnviron`.
2. **`NewProviderCommandContext` default-strips on unknown policy** — `provider_env.go:46-51`. Inverting the `switch` so that `EnvPolicyProviderChild` is the only stripping case makes unknown-policy behavior safe-by-default in the other direction (inherit instead of silently strip). Low priority while only two constants exist.
3. **HTTP proxy vars not in the strip list** — `provider_env.go:25-31`. `HTTP_PROXY`, `HTTPS_PROXY`, and lowercase variants are not stripped. No demonstrated loop risk today, but the comment promises "routing/proxy variables". Track in a follow-up if a concrete threat model emerges.

Issue #101, opened 2026-04-30
services: extensibility — plugin manifest overlays, runtime registration, systemd backend


chazmaniandinkle:
Part of #98. Depends on #100.

After Phase 2 (#100) establishes the kernel-side service control plane, Phase 3 makes services extensible (plugins can add their own) and cross-platform (Linux via systemd).

### 1. Plugin manifest overlays

Extend \`LoadManifest\` (\`internal/engine/node_manifest.go\`) to walk additional manifest files and overlay onto the base:

- Base: \`.cog/config/node/manifest.yaml\`
- Overlays: \`~/.cogos/plugins/*/manifest.yaml\`
- Conflict resolution: log + skip on duplicate \`name\` or \`port\` clash; first-loaded wins (base before plugins).

Plugin authors declare their services with the same \`ServiceDef\` schema. No code changes per plugin.

### 2. Runtime registration

\`\`\`
POST   /v1/services/register     { ServiceDef }
DELETE /v1/services/{name}       # only for runtime-registered services
\`\`\`

Persisted to \`~/.cogos/runtime-services.yaml\` (overlay loaded on next manifest reload). Idempotent on \`name\`. For plugins that register at install time without editing files manually.

### 3. SystemdSupervisor backend

Linux equivalent of \`LaunchctlController\`:

| Verb | Command |
|---|---|
| \`Start\` | \`systemctl --user start {unit}\` |
| \`Stop\` | \`systemctl --user stop {unit}\` |
| \`Restart\` | \`systemctl --user restart {unit}\` |
| \`Enable\` | \`systemctl --user enable {unit}\` |
| \`Disable\` | \`systemctl --user disable {unit}\` |
| \`Status\` | \`systemctl --user show {unit} --property=ActiveState,MainPID,...\` |

Add \`unit\` field to \`ServiceDef\` parallel to \`launchd\`. Selected at runtime via OS detection or explicit \`supervisor\` field.

## Out of scope

\`DirectSupervisor\` for dev-mode fork-children: defer to a separate small issue if/when needed. Container backends (Docker, podman): not currently a use case. Windows (\`sc.exe\`): defer.

## Acceptance

- A plugin manifest at \`~/.cogos/plugins/example/manifest.yaml\` is auto-loaded; its services appear in \`/v1/services\`.
- \`POST /v1/services/register\` creates a service that survives kernel restart.
[...]

Issue #98, opened 2026-04-30
services: kernel-managed services API + manifest-driven dashboard surface (umbrella)


chazmaniandinkle:
## Context

The kernel today has substantial service-management substrate that is not exposed through any HTTP surface:

- `NodeManifest` (`internal/engine/node_manifest.go`): declarative service catalog at `.cog/config/node/manifest.yaml`. Four services declared today: `kernel`, `mod3`, `gateway`, `cog-chat`. Per-service `ServiceDef` carries port, binary, command, health endpoint, restart policy, launchd label, `depends_on`, and a `consumers` list (files that reference this service's port, kept in sync by `cog node sync`).
- `NodeHealth.Probe()` (`internal/engine/node_probe.go:30`): concurrent `/health` probing of every declared service every cycle, 2 s timeout. Result feeds the `node` field of `GET /health`.
- `Reconcilable` interface (`pkg/reconcile/types.go`): k8s-shaped declarative reconciler. Autonomic ticker walks every Reconcilable every minute and emits `kernel.health.snapshot.v1` events to `bus_kernel_proprio`.

What's missing:

1. **No HTTP surface for the manifest.** Dashboards have to scrape the `node` field of `/health` and have no access to launchd labels, dependencies, or the `consumers` graph.
2. **No service control plane in the kernel.** launchctl integration today is just label declaration. The (now-archived) cogos-dev/desktop app shelled out to `launchctl` directly (`app.go:139-283`) as a workaround. Non-desktop clients (dashboard at :6931, CLI, future plugins) currently can't manage services at all.
3. **No `kind` discriminator.** Every `ServiceDef` is implicitly "managed". External services like Ollama / OpenClaw (when user-managed) have no way to be represented as observed-only.
4. **No plugin extensibility path.** The manifest is a single file with no overlay or runtime-registration mechanism.

## Principle

> The dashboard should be a live, manifest-driven mirror of the kernel API surface.

Two invariants follow:
1. No dashboard panel without a kernel route.
2. No kernel route without dashboard exposure.

`/v1/manifest` already enumerates all routes + MCP tools at runtime. The dashboard could derive panel discovery from it: custom renderers per known route family, generic API-explorer fallback for the rest. New kernel routes would auto-appear in the dashboard with no drift.

## Tracked sub-issues

Implementation phases:

- [x] **Phase 1**: read-only `/v1/services` API + `kind` discriminator (#99) — shipped via PR #108
[...]

PR #131, opened 2026-05-02
feat(services): Phase 2 — LaunchctlController + mutation endpoints (#100)

### LINKED ISSUE #100: services: LaunchctlController + service mutation endpoints (start/stop/restart/enable)
Part of #98. Depends on #99.

## Goal

Add the kernel's missing service-control plane. The desktop app currently shells out to \`launchctl\` directly (cogos-dev/desktop \`app.go:139-283\`). Lift that logic into the kernel so all clients (dashboard, CLI, future API consumers) manage services through one canonical endpoint.

## API surface

\`\`\`
POST /v1/services/{name}/start
POST /v1/services/{name}/stop
POST /v1/services/{name}/restart
POST /v1/services/{name}/enable      # launchctl load -w
POST /v1/services/{name}/disable     # launchctl unload -w
\`\`\`

Responses:
- \`200 OK\` with updated \`ServiceStatus\`.
- \`409 Conflict\` with \`{"error": "service not controllable", "kind": "observed"}\` when called on \`observed\` / \`external\`. The dashboard reads \`controllable\` from \`/v1/services\` and hides buttons accordingly. **The API never lies about what it can do.**

## Supervisor abstraction

Backends are pluggable so Linux can land later without re-architecting:

\`\`\`go
type ProcessSupervisor interface {
    Start(svc ServiceDef) error
    Stop(svc ServiceDef) error
    Restart(svc ServiceDef) error
    Enable(svc ServiceDef) error
    Disable(svc ServiceDef) error
    Status(svc ServiceDef) (SupervisorStatus, error)
}
\`\`\`

Implementations:
- \`LaunchctlController\` (macOS): **this issue**.
- \`SystemdSupervisor\` (Linux): Phase 3.
- \`DirectSupervisor\` (dev mode, fork children): Phase 3 if needed.
- \`ObserverSupervisor\`: read-only, returns \`ErrNotControllable\` on all mutations.

Selected via \`supervisor\` field on \`ServiceDef\`, defaulting per OS.

## LaunchctlController behavior

| Verb | Command |
|---|---|
| \`Start\` | \`launchctl kickstart -k gui/$UID/{label}\` |
| \`Stop\` | \`launchctl bootout gui/$UID/{label}\` |
| \`Restart\` | \`Stop\` then \`Start\` |
| \`Enable\` | \`launchctl load -w {plist_path}\` |
| \`Disable\` | \`launchctl unload -w {plist_path}\` |
| \`Status\` | parse \`launchctl print gui/$UID/{label}\` for PID, last exit code, state |

Plist paths derived from \`ServiceDef.launchd\` label + standard locations (\`~/Library/LaunchAgents/{label}.plist\`).

## Implementation sketch

- New \`internal/engine/supervisor.go\` with \`ProcessSupervisor\` interface + \`LaunchctlController\` impl.
- Wire mutation routes in \`serve_services.go\`.
- Surface \`pid\`, \`exit_code\`, \`running\` from \`Status()\` in the existing \`GET /v1/services\` projection.
- Test coverage with a mock \`ProcessSupervisor\` backend.

## Acceptance

- All five POST endpoints functional against real \`mod3\` service.
- \`409 Conflict\` on observed/external services.
- Manual smoke: dashboard POST \`/restart\` on mod3 takes effect, status reflects within next probe.

chazmaniandinkle:
## Summary

- Defines `ServiceSupervisor` interface (Start/Stop/Restart/Enable/Disable/Status) with `context.Context` on every method — ships two implementations on day one per the pinned operational-semantics spec on issue #100.
- `LaunchctlController` (darwin build-tag only): wraps `launchctl(1)` for `kind=managed` services with a launchd label. Implements all 12 operational-semantics edge cases from the pinned spec.
- `ObserverSupervisor` (all platforms): read-only stub returning `ErrNotControllable` for mutations; used by `kind=observed` and `kind=external` services.
- Five HTTP mutation endpoints: `POST /v1/services/{name}/start|stop|restart|enable|disable`.
- `EnableServiceControl` config gate (default `false`) in `kernel.yaml`, matching the `EnableSkillExec` pattern.
- 23 test cases covering the pinned spec as a test-case checklist.

## Changes

**New files:**
- `internal/engine/service_supervisor.go` — interface, `ObserverSupervisor`, `ErrNotControllable`, `supervisorFor()` routing helper
- `internal/engine/service_supervisor_launchctl.go` — `LaunchctlController` (darwin only)
- `internal/engine/service_supervisor_stub.go` — non-darwin stub (`LaunchctlController = ObserverSupervisor`)
- `internal/engine/serve_services_mutations_test.go` — 23 tests

**Modified:**
- `internal/engine/serve_services.go` — mutation handlers, `registerServiceMutationRoutes`, `dispatchMutation` core, `serviceMutationResponse` type
- `internal/engine/serve.go` — `serviceSupervisor` field, `SetServiceSupervisor()`, registers mutation routes at startup
- `internal/engine/config.go` — `EnableServiceControl bool` field + YAML key `enable_service_control`

## Test plan

- [x] `go test ./internal/engine/ -count=1 -timeout 120s` passes (all 33 service tests + full suite)
- [x] Gate: 403 on all verbs when `EnableServiceControl=false`
- [x] Not-found: 404 for unknown service name; 404 for nil manifest
- [x] Not-controllable: 409 for `kind=observed` and `kind=external`
- [x] Happy path: 200 for all verbs on managed service via `stubSupervisor`
- [x] Stop includes `Stopping=true` in status
- [x] Start idempotent on already-running (200, not 409)
- [x] Stop idempotent on already-stopped (200)
- [x] Enable/Disable idempotent (double-call → 200 both times)
- [x] Error shape: `success=false`, `error` non-empty, `launchctl_exit_code` present
- [x] `ObserverSupervisor` unit: all mutations — `ErrNotControllable`; `Status` — valid snapshot
- [x] `supervisorFor` routing: managed+controller — controller; nil — observer fallback; observed/external — observer always
- [ ] Manual: `curl -X POST localhost:6931/v1/services/kernel/start` with `enable_service_control: true` in `kernel.yaml`

## Out of scope

- Phase 3 (#101): plugin overlays, runtime registration, `SystemdSupervisor`, `DirectSupervisor`.
- `?wait=true` query parameter (spec calls it optional; left for follow-up).
- Bus event emission on mutation (can be added alongside bus-native agent work).
- `LaunchctlController` integration tests — require a live launchd instance; excluded from CI.

## Operational-semantics deviations / open questions

None. The implementation follows the pinned spec exactly. One point for reviewers:

> **Restart serialisation**: The per-service mutex in `LaunchctlController.Restart` serialises concurrent calls. The second caller acquires the same `sync.Mutex` and re-executes stop+start (idempotent). The spec says "return same response from both" — both callers get the result of their own restart attempt, which is functionally equivalent given stop+start idempotency. If strict shared-result semantics are required, a `singleflight`-shaped promotion can be layered on; flagging for review.

Closes #100. Part of #98.

### DIFF
--- a/internal/engine/config.go
+++ b/internal/engine/config.go
@@ -124,6 +124,13 @@ type Config struct {
 	// (`cog skill exec`) is unaffected — it runs in the user's own session.
 	EnableSkillExec bool
 
+	// EnableServiceControl gates the service mutation endpoints:
+	// POST /v1/services/{name}/start|stop|restart|enable|disable.
+	// Default false: service mutations via HTTP are disabled by default because
+	// any local process on the same host could otherwise trigger launchctl
+	// operations. Set enable_service_control: true in kernel.yaml to opt in.
+	EnableServiceControl bool
+
 	// DigestPaths maps stream tailer adapter names to JSONL file/directory paths.
 	// Empty map means external digestion is disabled.
 	DigestPaths map[string]string
@@ -166,6 +173,7 @@ type kernelConfigSection struct {
 	OllamaEmbedModel      string            `yaml:"ollama_embed_model"`
 	ToolCallValidation    *bool             `yaml:"tool_call_validation_enabled"`
 	EnableSkillExec       *bool             `yaml:"enable_skill_exec"`
+	EnableServiceControl  *bool             `yaml:"enable_service_control"`
 	LocalModel            string            `yaml:"local_model"`
 	DigestPaths           map[string]string `yaml:"digest_paths"`
 	KernelLogPath         string            `yaml:"kernel_log_path"`
@@ -290,6 +298,9 @@ func applyKernelSection(cfg *Config, s kernelConfigSection) {
 	}
 	if s.EnableSkillExec != nil {
 		cfg.EnableSkillExec = *s.EnableSkillExec
+	}
+	if s.EnableServiceControl != nil {
+		cfg.EnableServiceControl = *s.EnableServiceControl
 	}
 	if s.LocalModel != "" {
 		cfg.LocalModel = s.LocalModel
--- a/internal/engine/serve.go
+++ b/internal/engine/serve.go
@@ -62,6 +62,7 @@ type Server struct {
 	nucleus         *Nucleus
 	process         *Process
 	router          Router // nil until SetRouter is called
+	serviceSupervisor ServiceSupervisor // nil until SetServiceSupervisor; defaults to ObserverSupervisor
 	srv             *http.Server
 	debug           debugStore      // captures last request pipeline state
 	attentionLog    *attentionLog   // per-server log (avoids global write race)
@@ -187,9 +188,10 @@ func NewServer(cfg *Config, nucleus *Nucleus, process *Process) *Server {
 	// ambient-awareness loop; Phase 1A populates channel.<sid>.activity).
 	s.registerPeerAwarenessRoutes(mux)
 
-	// Read-only services API: GET /v1/services and GET /v1/services/{name}.
-	// Phase 2 will add mutations (start/stop/restart).
+	// Services API: GET /v1/services and GET /v1/services/{name} (Phase 1);
+	// POST /v1/services/{name}/{action} mutations (Phase 2).
 	s.registerServiceRoutes(mux)
+	s.registerServiceMutationRoutes(mux)
 
 	// Replay bus_sessions + bus_handoffs into the in-memory registries so
 	// the kernel starts with an accurate derived view. Bus is authoritative
@@ -223,6 +225,13 @@ func NewServer(cfg *Config, nucleus *Nucleus, process *Process) *Server {
 // SetRouter wires an inference Router into the server.
 func (s *Server) SetRouter(r Router) {
 	s.router = r
+}
+
+// SetServiceSupervisor wires a ServiceSupervisor into the server for
+// service mutation endpoints (start/stop/restart/enable/disable).
+// If not called, all mutation endpoints use ObserverSupervisor (read-only).
+func (s *Server) SetServiceSupervisor(sup ServiceSupervisor) {
+	s.serviceSupervisor = sup
 }
 
 // SetAgentController wires a live AgentController into the server so the
--- a/internal/engine/serve_services.go
+++ b/internal/engine/serve_services.go
@@ -1,23 +1,30 @@
-// serve_services.go — read-only /v1/services API.
+// serve_services.go — /v1/services API (read-only Phase 1 + mutations Phase 2).
 //
 // Routes:
 //
-//	GET /v1/services         — list all declared services with live health
-//	GET /v1/services/{name}  — single service projection
+//	GET  /v1/services                    — list all declared services with live health
+//	GET  /v1/services/{name}             — single service projection
+//	POST /v1/services/{name}/start       — start a managed service
+//	POST /v1/services/{name}/stop        — stop a managed service
+//	POST /v1/services/{name}/restart     — restart a managed service
+//	POST /v1/services/{name}/enable      — enable a managed service (boot persistence)
+//	POST /v1/services/{name}/disable     — disable a managed service (remove boot persistence)
 //
 // Design:
-//   - Projects from two read sources: the parsed *NodeManifest (static
-//     declaration) and *NodeHealth (last probe snapshot). Both are held on
-//     *Process and accessed through NodeManifest() / NodeHealth() accessors.
-//   - Read-only. Mutations (start / stop / restart) are Phase 2.
+//   - GET routes project from *NodeManifest (static declaration) and *NodeHealth
+//     (last probe snapshot). Both are held on *Process.
+//   - Mutation routes are gated by Config.EnableServiceControl (default false).
+//     Each mutation calls s.serviceSupervisor; if nil, defaults to ObserverSupervisor.
 //   - "kind" defaults to "managed" when omitted from the manifest.
 //   - "controllable" is false for observed and external services.
+//   - Mutations on observed/external services return 409 with ErrNotControllable.
 //
 // Registered via s.route() so the routes auto-appear in GET /v1/manifest.
 package engine
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 	"sort"
 	"strings"
@@ -52,6 +59,16 @@ type serviceHealthV struct {
 func (s *Server) registerServiceRoutes(mux *http.ServeMux) {
 	s.route(mux, "GET /v1/services", s.handleServicesList)
 	s.route(mux, "GET /v1/services/{name}", s.handleServicesGet)
+}
+
+// registerServiceMutationRoutes wires the Phase 2 mutation endpoints.
+// All are gated by Config.EnableServiceControl.
+func (s *Server) registerServiceMutationRoutes(mux *http.ServeMux) {
+	s.route(mux, "POST /v1/services/{name}/start", s.handleServiceStart)
+	s.route(mux, "POST /v1/services/{name}/stop", s.handleServiceStop)
+	s.route(mux, "POST /v1/services/{name}/restart", s.handleServiceRestart)
+	s.route(mux, "POST /v1/services/{name}/enable", s.handleServiceEnable)
+	s.route(mux, "POST /v1/services/{name}/disable", s.handleServiceDisable)
 }
 
 // handleServicesList returns all declared services sorted by name.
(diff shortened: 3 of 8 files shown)


chazmaniandinkle (commented):
## Verdict

**minor-follow-ups** — ship after the Restart cache fields are cleaned up. No correctness blockers; one dead-code bug in the serialisation mechanism and one spec deviation (503 vs 500 for transient launchd errors) that belong in follow-up issues rather than blocking merge.

---

## Acceptance — 12 Operational-Semantics Cases

Mapping each spec item from the pinned comment on #100:

| # | Spec item | Test coverage | Notes |
|---|-----------|--------------|-------|
| 1 | `start()` idempotent — already-running → 200 | `TestServiceMutation_Start_IdempotentWhenAlreadyRunning` ✓ | — |
| 2 | `restart()` mid-transition — per-service mutex, concurrent callers serialise | No unit test for concurrent case | Logic present but cache fields are dead code (see Code section). Single-threaded path tested via `TestServiceMutation_Restart_Success`. |
| 3 | Status freshness ≤ ~2s; optional `?wait=true` | Not tested; explicitly deferred (OOS) | PR body calls this out. Acceptable. |
| 4 | Exit-code mapping — 400/409/503/500; raw exit code in body | `TestServiceMutation_ErrorShape` covers `launchctl_exit_code` field; 503 mapping absent | See Code section — no sentinel for transient launchd errors; they fall through to 500. |
| 5 | Missing plist → `controllable: false`; don't auto-load | No test | `Start()` _does_ attempt auto-load (`launchctl load <path>`) when not registered. Diverges from spec's "don't auto-load" intent. Launchd will error if plist absent; error surfaces as 500. Follow-up worthy. |
| 6 | Per-service lock + in-flight flag; two concurrent restarts → serialise, return same response | No concurrent test | Cache fields are dead code (below). Serialisation works but "return same response from both" is not achieved. |
| 7 | `controllable` reliability when launchd forgets — `start()` re-registers via `launchctl load` then kickstarts | No dedicated test; impl correct | `Start()` calls `load` then `kickstart` when `!st.LaunchdRegistered`. ✓ |
| 8 | `disable()` semantics — `launchctl unload -w`; idempotent | `TestServiceMutation_Disable_Idempotent`, `TestServiceMutation_Disable_Success` ✓ | — |
| 9 | `stop()` graceful — SIGTERM 5s timeout, then SIGKILL; return 200 + `stopping: true` immediately | `TestServiceMutation_Stop_ReturnsStopping` ✓ (at HTTP layer) | SIGTERM/SIGKILL delegated to launchd via `launchctl stop`; integration test excluded from CI per PR body. Acceptable for Phase 2. |
| 10 | Service-not-found → 404; distinct from 400 | `TestServiceMutation_NotFound` (all verbs) ✓ | — |
| 11 | launchd authoritative for `running`; `/health` bonus signal for `healthy` | Design correct; no explicit test | `Status()` uses `launchctl list` exclusively. `projectService()` uses health probe for GET projection only (separate surface). ✓ |
| 12 | Error response shape: `{ success, error, status, launchctl_exit_code }` | `TestServiceMutation_ErrorShape`, `TestServiceMutation_ResponseShapeInvariants` ✓ | — |

**Gaps**: items 2/6 (concurrent restart test), item 4 (503 sentinel), item 5 (missing plist behaviour). Items 3 and 9 are explicitly deferred.

---

## Code

### Restart — dead cache fields (`service_supervisor_launchctl.go:57-62`, `168-198`)

This is the deviation called out in the PR body. The analysis:

`serviceRestartLock` holds `inFlight bool`, `result *ServiceStatus`, `err error`. The intent is: goroutine B arriving while A is in-flight reads A's cached result rather than re-executing.

The implementation does not achieve this. Because the lock is a `sync.Mutex` (not a `sync.RWMutex` + channel), goroutine B blocks at `lock.mu.Lock()` until A calls `defer lock.mu.Unlock()`. By the time B acquires the lock, A has already set `lock.inFlight = false`. B's `if lock.inFlight` check evaluates false; B falls through and re-executes the full stop+start.

`lock.result` and `lock.err` are written on every path but read on no reachable path. The comment on line 172 acknowledges the real behaviour ("we simply execute the restart again") but leaves the dead cache fields in place.

**Correctness impact**: zero — re-executing stop+start is safe and idempotent. The spec's "return same response from both" is not met, but the PR body acknowledges this and flags it as an acceptable deviation given idempotency.

**What the dead code signals**: the original design was singleflight-shaped (shared result). The impl fell back to re-execution but didn't remove the cache scaffolding. This should be cleaned up before it confuses the next reader.

**Recommended follow-up**: either simplify `serviceRestartLock` to just `sync.Mutex` (3 fields → 1), or replace with `golang.org/x/sync/singleflight` if shared-result semantics become a requirement. File as a separate issue.

### Redundant empty-label guard in `Start()` (`service_supervisor_launchctl.go:87-100`)

Line 87 assigns `label := def.Launchd` and line 88 returns an error if `label == ""`. The inner check at line 94 (`if def.Launchd == ""`) is therefore unreachable — `label` is guaranteed non-empty at that point. This is a nit but it's confusing: the inner branch returns a soft `{ running: false }` status with no error, while the outer guard returns a hard error. The divergence in error handling between these two identical conditions suggests copy-paste drift. The inner branch (lines 94-101) is dead code and should be removed.

### 503 mapping absent (`serve_services.go:297-311`)

The `dispatchMutation` switch maps `ErrNotControllable → 409`, `ErrServiceNotFound → 404`, `ctx.Err → 400`, and everything else → 500. The launchctl file header documents that exit code 125 / "service failed" should map to 503, but no sentinel error type carries that signal. Transient launchd failures surface as generic `fmt.Errorf(...)` wrapped around an `exec.ExitError`, which falls to 500. The `launchctl_exit_code` field in the response body lets clients distinguish this case themselves, which partially mitigates the gap. Still, the spec says 503 for transient failures. Worth a follow-up issue.

### `errors.Is(err, r.Context().Err())` context check (`serve_services.go:307-310`)

When the request context is not cancelled, `r.Context().Err()` returns `nil`. `errors.Is(someErr, nil)` returns false for any non-nil error, so there is no false-positive risk. When the context is cancelled, the call correctly maps to 400. This pattern is valid; no issue here.

### Gate parity with `EnableSkillExec`

`requireServiceControl` uses `json.Encoder` with `Content-Type: application/json` set explicitly. `requireSkillExec` in `serve_skills.go:264` uses `http.Error()` which does not set `Content-Type: application/json`. The service implementation is the better pattern. This is a pre-existing gap in the skill gate, not a regression introduced here.

---

## Tests

23 new tests in `serve_services_mutations_test.go`. All pass (`go test ./internal/engine/ -count=1`, 7.3 s).

**Adequate for what they cover.** The happy-path and gate tests are well-structured; the `stubSupervisor` is clean and reusable.

**Gaps relative to the spec checklist:**

1. No concurrent-restart test. Spec items 2 and 6 require two goroutines hitting `Restart` simultaneously. The absence of a race-detector hit here doesn't prove correctness — it just means the race hasn't been caught. Given that the cache fields are dead code anyway, a concurrent test would expose the "re-executes rather than shares result" behaviour and document it explicitly. This is a test-for-the-follow-up more than a block.

2. `TestServiceMutation_GateDisabled` checks `strings.Contains(resp.Error, "disabled")`. The gate writes a raw `map[string]string` with `"error": "disabled"`. When decoded into `serviceMutationResponse`, `resp.Error` = `"disabled"` and `resp.Action` = `""`. The test does not assert `resp.Action` or `resp.ServiceName` — reasonable given the gate fires before those are resolved.

3. No test for `Start()` when service has a launchd label but the plist file is absent on disk (spec item 5). Not blocking, but worth noting.

---

## Follow-ups

### 1. Clean up `serviceRestartLock` dead cache fields

**File**: `internal/engine/service_supervisor_launchctl.go:57-62`, `168-178`

Remove `inFlight`, `result`, `err` from `serviceRestartLock` (or replace the whole struct with a plain `sync.Mutex`). If shared-result singleflight semantics are desired later, replace with `golang.org/x/sync/singleflight`. The current struct misleads readers into thinking shared-result is implemented when it is not.

### 2. Define a sentinel error type for transient launchd failures and map to 503

The spec documents exit 125 → 503. Currently all non-sentinel launchctl errors → 500. Adding an `ErrLaunchdTransient` sentinel in `service_supervisor_launchctl.go` and a corresponding case in `dispatchMutation` would close this gap.

### 3. Resolve spec item 5 — missing plist behaviour

`Start()` currently calls `launchctl load <plistPath>` for unregistered services regardless of whether the plist file exists. The spec says "missing plist → controllable: false; don't auto-load." Either add a `os.Stat` check before `load`, or document the deviation as intentional. As-is, a missing plist surfaces as a 500 with a launchctl error body, not a clean `controllable: false`.

chazmaniandinkle:
## Fix-up commit summary

Two spec divergences from the pinned operational-semantics spec (#100) have been addressed. The Restart dead-cache-fields cleanup is explicitly deferred to #139.

---

### Fix 1 — 503 for transient launchd errors (spec item 4)

**Commit:** `fix(services): map launchctl exit 125 to HTTP 503 (spec item 4)`

- Added `ErrLaunchctlTransient` sentinel in `service_supervisor.go` with a comment tying it to the exit-125 / "service failed" case documented in the spec.
- Added `wrapTransientErr(err, exitCode)` helper in `service_supervisor_launchctl.go`; applied at the `kickstart` call site in `Start` and at the `stop` call site in `Stop`. When `launchctl` exits 125 the error is wrapped with `%w: %w` so `errors.Is` traversal finds the sentinel.
- Added the `ErrLaunchctlTransient → 503` case to `dispatchMutation`'s error-to-status switch in `serve_services.go`, positioned above the context-cancel catch-all.
- Added `TestServiceMutation_TransientLaunchd503`: stubs `startErr` as `fmt.Errorf("%w: ...", ErrLaunchctlTransient)` with `LaunchctlExitCode: 125`, asserts HTTP 503 and `launchctl_exit_code=125` in the response body.

---

### Fix 2 — Missing plist → `ErrNotControllable` (spec item 5)

**Commit:** `fix(services): stat plist before launchctl load; missing plist → ErrNotControllable (spec item 5)`

- In `LaunchctlController.Start`, before `launchctl load`, `os.Stat` the resolved plist path. If `os.IsNotExist`, return `ErrNotControllable` (wrapped with `%w` + a message including the path) with a status of `{ running: false, launchd_registered: false }`. No launchctl subprocess is spawned.
- This maps to HTTP 409 through the existing `ErrNotControllable` case in `dispatchMutation` — the same shape `ObserverSupervisor` returns for kind=observed services.
- Added `TestLaunchctlController_Start_MissingPlist` (darwin-only, `//go:build darwin`) in a new `service_supervisor_launchctl_test.go`: creates a `LaunchctlController`, builds a `ServiceDef` with a label guaranteed to be absent from launchd and whose plist path does not exist, calls `Start`, and asserts `errors.Is(err, ErrNotControllable)` with `running=false, launchd_registered=false`.

---

### Restart cleanup

The dead `inFlight`/`result`/`err` cache fields in `serviceRestartLock` are **not** touched here — that cleanup is tracked at #139.

---

**All 25 engine service tests pass** (`go test ./internal/engine/ -count=1 -timeout 120s`).

PR #145, opened 2026-05-02
fix(engine): raise local-provider HTTP timeout to cover Ollama cold-load

### LINKED ISSUE #144: Ollama model lifetime not kernel-managed: keep_alive lost on first non-kernel request, cold-start retries thrash memory
## Summary

The kernel's local-inference path does not enforce model lifetime against Ollama. The model (`gemma4:e4b`, ~11 GB VRAM) only stays pinned as long as the *most recent* request to Ollama specified `keep_alive: -1` — and Ollama applies "last request wins" to that field. Any other client (an `ollama run`, embed-server, dashboard `/v1` fallback, manual experiment) that hits the API without setting `keep_alive` resets the timer to the server default (5 min). When the model is then evicted and the next dispatch arrives, the kernel's 90 s `localHarnessCycleTimeout` is shorter than gemma's ~110 s cold-start, so the load is aborted mid-flight and retried — visible memory pressure ensues.

This was diagnosed against a live host where `~/.cog/var/logs/serve.log` and Ollama's `server.log` both showed three back-to-back load attempts (10:35 / 10:49 / 11:03 today), the first two failing with `client connection closed before server finished loading, aborting load`.

The audit report `cog://workspace/.cog/scratch/audit-inference-paths/REPORT.md` already calls this out: *"Kernel-managed lifetime? No. Ollama's `keep_alive=-1` is set in `providers.local.yaml:13`."* — but `providers.local.yaml:13` is a YAML *comment*, not config the kernel reads; lifetime is fully external.

## Evidence

```
$ curl -s http://localhost:11434/api/ps
{"models":[{"name":"gemma4:e4b",...,"expires_at":"2318-08-12T12:51:00...","size_vram":11074136544,...}]}
```

`expires_at: 2318` was set by *some* prior request (likely a manual `ollama run --keepalive -1` or a kernel call that did pass it), but it survives only until the next request that doesn't.

`~/.ollama/logs/server.log` (today):
```
10:35:23  WARN  client connection closed before server finished loading, aborting load
10:35:23  ERROR error loading llama server: timed out waiting for llama runner to start: context canceled
10:49:24  INFO  loading model "model layers"=43 requested=-1
10:49:36  WARN  client connection closed before server finished loading, aborting load
11:03:36  INFO  loading model ... [eventually succeeds]
```

Three loads in ~30 minutes for a model that should be permanently resident.

## Root causes

1. **Per-request `keep_alive` is fragile.** The kernel sends inference requests via `cogos/internal/engine/provider_ollama.go` → `POST /api/chat`, but `keep_alive` is not unconditionally `-1` in the request body. Even if it were, any non-kernel client overrides it on the next call.

2. **Kernel cold-start timeout < gemma cold-start.** `localHarnessCycleTimeout` defaults to 90 s; gemma4:e4b consistently takes ~110 s to load on cold cache. The HTTP client gives up before Ollama finishes, Ollama logs `client connection closed`, the next dispatch retries, and Ollama begins another load. Load amplification under memory pressure.

3. **Lifetime is external.** Even with `OLLAMA_KEEP_ALIVE=-1` set on the Ollama side, the kernel has no contract with Ollama about model residency — it can't observe eviction, prevent eviction, or warm the model proactively. Per the audit's Tier 3 plan this is the long-term fix; in the meantime the env-var workaround is the right floor.

4. **Duplicate Ollama daemon noise.** On at least one host, `~/Library/LaunchAgents/homebrew.mxcl.ollama.plist` is enabled alongside Ollama.app, fails to bind 11434 (`exit code 78`, EX_CONFIG), and is held in `spawn scheduled` state by `KeepAlive=true`. Doesn't allocate the model but spams logs. Worth documenting as a known foot-gun in setup docs.

### Short-term (operator-side, ~10 min)

Document and recommend setting these on the Ollama process:

```
OLLAMA_KEEP_ALIVE=-1
OLLAMA_MAX_LOADED_MODELS=1
OLLAMA_NUM_PARALLEL=1
```

Server-side defaults; no client can override them downward. Add to `providers.local.yaml` operator notes.

### Medium-term (kernel-side, scoped fix)

- `provider_ollama.go`: unconditionally include `"keep_alive": -1` in every `/api/chat` request body when the provider has `is_local: true` or matches the resident-model alias. Cheap defense against any non-kernel client that just touched Ollama.
- Bump `localHarnessCycleTimeout` default from 90 s → 180 s, OR carve out a separate cold-load timeout (~180 s) distinct from the steady-state inference timeout. The 90 s value was sized for a warm runner; cold is a different regime.
- Optional: kernel-side warmup ping on startup (`POST /api/chat` with `messages: []` and `keep_alive: -1`) to deterministically claim the residency slot rather than waiting for first user dispatch.

### Long-term (architectural)

Tier 3 from the inference-paths audit: kernel-supervised in-host inference (`mlx_lm.server` child under `ProcessManager`) so model lifetime is kernel-owned and observable, not negotiated through Ollama's "last request wins" semantics. Tracked separately.

## Out of scope

- Metal residency-set `keep_alive = 180s` (hardcoded in the `ollama runner` Go source) — not configurable from outside Ollama; affects GPU page wiredness, not model presence. Mention in docs only.
- System-level memory pressure (host had 9.7 GB swap in use during diagnosis) — independent of kernel.

## References

- Audit: `cog://workspace/.cog/scratch/audit-inference-paths/REPORT.md`
- Existing memory note: `feedback_substrate_eval_constraints` (90 s timeout < 110 s cold-start, Ollama 5-min default eviction)
- Existing memory note: `feedback_ollama_single_thread_constraint` ("last request wins" on `keep_alive`)

chazmaniandinkle:
Closes #144 (in part — operator-side `OLLAMA_KEEP_ALIVE=-1` guidance is separate, applied out-of-band on the diagnosis host).

## What

`buildLocalProvider` (`local_llm.go:130`) was hardcoding `Timeout: 120` on the per-dispatch `ProviderConfig`. The resulting `OllamaProvider` constructed `&http.Client{Timeout: 120 * time.Second}`. Two problems:

1. 120 s is below gemma4:e4b's cold-load on Apple Silicon (~110 s clean, ~150–180 s under memory pressure). The HTTP client closes the connection mid-load → Ollama logs `client connection closed before server finished loading, aborting load` → next dispatch triggers a fresh load → amplified memory spike.
2. The hardcoded literal silently shadowed the operator's `timeout: 300` in `providers.local.yaml` — that config was effectively dead code on this path.

This bumps the literal to 300 s, matching what operators already had configured.

## Why this is the right scope

Threading the configured timeout through `buildLocalProvider` (so YAML actually wins) is the principled fix, but `buildLocalProvider`'s signature only receives `target LocalLLMTarget` and `model string` — it has no access to the workspace provider config. Plumbing it requires changing both call sites in `local_agent_harness.go` (lines 412 and 819) and threading config through. Worth doing, but separate. This PR is the one-line stop-gap that immediately ends the reload thrash observed in #144.

## Evidence from the diagnosis host

Server log on the day this was filed:

```
10:35:23  WARN  client connection closed before server finished loading, aborting load
10:35:23  ERROR error loading llama server: timed out waiting for llama runner to start: context canceled
10:49:24  INFO  loading model "model layers"=43 requested=-1
10:49:36  WARN  client connection closed before server finished loading, aborting load
11:03:36  INFO  loading model ... [eventually succeeds]
```

The 11–14 second gap between load-start and "client closed" matches a 120 s timeout firing partway into a cold-load that started ~108 s earlier in a previous (failed) request and was still in-flight when the next dispatch arrived.

## Test

`go build ./...` ✓
`go test ./internal/engine/...` ✓ (7.069s)

No new tests — the change is a single-literal value bump and existing OllamaProvider tests cover the construction path.

## Follow-up tracked separately

- Thread `cfg.Timeout` from workspace provider config through `buildLocalProvider` so YAML config isn't shadowed.
- Long-term: kernel-supervised in-host inference (audit Tier 3) so model lifetime is observable to the kernel, not negotiated through Ollama's "last request wins" semantics.

### DIFF
--- a/internal/engine/local_llm.go
+++ b/internal/engine/local_llm.go
@@ -127,7 +127,11 @@ func buildLocalProvider(target LocalLLMTarget, model string) Provider {
 	cfg := ProviderConfig{
 		Endpoint: target.BaseURL,
 		Model:    model,
-		Timeout:  120,
+		// 300s covers cold-load of an 8B-class quantized model on Apple Silicon
+		// under memory pressure (~110s clean, ~150-180s under load). 120s
+		// closed the connection mid-load, aborting Ollama's load and forcing
+		// a retry on the next dispatch. See cogos-dev/cogos#144.
+		Timeout: 300,
 	}
 	switch target.Backend {
 	case LocalLLMBackendOllama:


chazmaniandinkle:
Superseded by #146, which carries the same fix wrapped in proper config threading: the 300 s value lives in `localProviderDefaultTimeoutSec` as the fallback when no provider config is available, and `resolveLocalProviderTimeout` reads `providers(.local).yaml` so operator-configured timeouts actually take effect (the deeper bug this one didn't address). Closing this branch in favor of the consolidated fix.

Issue #167, opened 2026-05-03
kernel: implement URIRegistry + alias subsystem (`pkg/filelock`, `pkg/alias`, CLI)


chazmaniandinkle:
## Context

The kernel has a `URIRegistry` extension point at `internal/engine/uri_v2_stub.go:28` that is `nil` at runtime today. MCP at `internal/engine/mcp_server.go:815-853` falls back to legacy local-only resolution. This issue implements the registry and the alias subsystem on top of it.

## Architecture

Three node-config files at `~/.cog/node/`:
- `identity.yaml` — node identity (already exists)
- `global.yaml` — workspace registry (post-PR #163; required prerequisite)
- `aliases.yaml` — alias map (this issue introduces)

## v0 schema

```yaml
# ~/.cog/node/aliases.yaml
version: "1.0"
aliases:
  cog:      cog-workspace
  kernel:   cogos-dev/cogos
  m3:       cogos-dev/mod3
  cogfield: cogos-dev/cogfield
  docs:     cogos-dev/docs
```

Schema rules:

- Alias name: `^[a-z][a-z0-9_-]{0,30}$`
- Target: literal workspace name registered in `global.yaml`. Not another alias.
- Optional `node:` and `description:` per entry
- No `trust`/`sync`/`pin`/`branch` fields (those live elsewhere — pin records, workspace registration)

## Resolver precedence

Per ADR-067's amended grammar:
1. Parse URI; identify slash count
2. `cog:path` → projection lookup in current workspace (no alias check)
3. `cog://authority/path` → authority lookup:
   - Alias map first (`aliases.yaml`)
   - Workspace registry second (`global.yaml`)
   - `ErrUnknownAuthority` if neither
4. Workspace name → filesystem root via `global.yaml`
5. Path resolution via existing projections logic
6. Digest verification if `?digest=` present (fail-closed per ADR-067 amendment)

## Storage discipline

- Aliases never serialize to cogdocs. Kernel rewrites alias→canonical on write.
- Display path: kernel may elide canonical→alias for human-readable rendering (foveated context, peer-awareness packets, search results).
- Logs and error messages always include canonical form; alias in parens for debugging context.

## v0 implementation slice

**New files:**

- `pkg/filelock/lock.go` (~60 lines) — `syscall.Flock` wrapper with 2s timeout → `ErrLockTimeout`. Wraps writes to `aliases.yaml` and `global.yaml`. Pattern: lock → read → mutate → atomic-write → unlock.
[...]

PR #168, opened 2026-05-03
feat(pin): inter-workspace pin Reconcilable + CLI

chazmaniandinkle:
## Summary

- Implements the seven-method `Reconcilable` interface for inter-workspace pin records at `<workspace>/.cog/pins/<target>.yaml` (`internal/providers/pin/pin.go`, ~300 lines).
- Registers the pin provider in both the daemon proprioception health block (`internal/providers/daemon/daemon.go`) and the CLI binary (`cmd_pin.go` init), so pin drift surfaces in foveated context automatically.
- Adds `cog pin list|add|remove|verify|bump` CLI (`cmd_pin.go`).

## Key design decisions

- **Pins are source-side.** No changes to `global.yaml` workspace entries or the `trust:` field — those are membership facts. `sync:`, `branch:`, `pin:` are relationship fields on the pin record only.
- **Read-only default.** `ApplyPlan` refuses to write-back for `sync: read-only` pins; explicit `cog pin bump` is the only path to a ref update.
- **Sibling-checkout fallback (v0).** `FetchLive` resolves target workspace paths via local git checkout inspection. A clear `TODO(#167)` marks the `URIRegistry.Resolve` integration point. The follow-up PR wires URIRegistry after #169 lands. This handles all current use cases — federation is not yet deployed, so the local-checkout path covers every real pin today.
- **Digest mismatch → red.** When a pin record declares a `pin.digest`, a mismatch with the live digest drives Health to Degraded/OutOfSync.

## Rebase note (post-#170)

Rebased cleanly against `upstream/main` after PR #170 (resolver consolidation) merged. The #170 changes to `cog.go` (`cogURIPattern`, `resolveURI`, `validateCogdoc*`) are in entirely different regions from the `"pin":` dispatch case added here — zero conflicts. The daemon test was updated from 9→10 expected providers to reflect the new pin registration.

`ErrUnknownAuthority` / `ErrDigestNotVerified` from `internal/engine/uri.go` (added in #170) are not imported by the pin package yet — the stub's error path uses a plain `fmt.Errorf`. The URIRegistry wire-up follow-up will wrap with `ErrUnknownAuthority` at that point.

## URIRegistry wire-up

The `TODO(#167)` wire-up is tracked in a follow-up issue. It will land in a small PR after #169 (URIRegistry) merges.

## Files touched

| Category | Files |
|----------|-------|
| New pin package | `internal/providers/pin/pin.go`, `internal/providers/pin/pin_test.go` |
| Health registration | `internal/providers/daemon/daemon.go`, `internal/providers/daemon/daemon_test.go` |
| CLI + dispatch | `cmd_pin.go`, `cog.go` |

## Test results

```
ok  github.com/cogos-dev/cogos/internal/providers/pin     (12/12 tests pass)
ok  github.com/cogos-dev/cogos/internal/providers/daemon  (4/4 tests pass)
ok  github.com/cogos-dev/cogos                            full suite green
ok  github.com/cogos-dev/cogos/internal/engine            full suite green
```

### DIFF
--- a/cmd_pin.go
+++ b/cmd_pin.go
@@ -182,10 +182,8 @@ func cmdPinRemove(args []string) error {
 // cmdPinVerify resolves live HEAD for each pin and reports drift.
 // Optional argument: specific target to verify. Omit to verify all.
 //
-// The output is driven entirely from the ComputePlan result — the same plan
-// that Health() reads. This ensures that digest mismatches, unreachable targets,
-// and ref drift are all reflected correctly, matching the three-axis health
-// classification rather than a second ad-hoc ref comparison.
+// This is a thin shim over pin.RunVerify so the core logic is testable
+// independently of workspace resolution and os.Stdout.
 func cmdPinVerify(args []string) error {
 	var filterTarget string
 	if len(args) > 0 && !strings.HasPrefix(args[0], "-") {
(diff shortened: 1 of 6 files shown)


chazmaniandinkle:
## Fix commits — Codex review findings addressed

Three blocking findings from the Codex review (PR #168, session 019def74) have been resolved. Each commit is surgical: no subsystem redesign, no schema changes, no pin record moves.

---

### Bug 1 — Daemon health cannot surface drift

**Codex finding:** `internal/providers/daemon/daemon.go:86-109` calls `LoadConfig` then `Health`. `LoadConfig` sets `root` but does not run `FetchLive`/`ComputePlan`, so `pinStates` is always empty. `Health` with empty `pinStates` reports "no pins declared" (green) regardless of declared pin files.

**Fix commit:** `936f5e6` — `fix(daemon): call full Reconcile cycle before Health() in pin provider`

`pinProvider.Health()` now calls `impl.Reconcile(ctx, root)` (full `LoadConfig → FetchLive → ComputePlan → ApplyPlan` cycle) before reading `impl.Health()`. A 10 s context timeout prevents blocking the proprioception tick. A 30 s staleness gate reuses the last result for rapid repeated calls.

**New test:** `TestPinProvider_DaemonHealth_SurfacesDrift` in `internal/providers/daemon/daemon_test.go`

---

### Bug 2 — Digest pins fail OPEN (contradicts ADR-067 fail-closed amendment)

**Codex finding:** `pin.go:200-208` local resolver returns empty digest. `ComputePlan` at `pin.go:428-452` only compares digests when `rec.Pin.Digest != "" && lr.Digest != ""` — so a declared digest with no live digest goes green.

**Fix commit:** `f21eb2f` — `fix(pin): digest declared + empty live resolver = RED (fail-closed, ADR-067)`

`ComputePlan` now handles three explicit cases:
1. `pinned == ""` → no digest check (digest is optional)
2. `pinned != ""`, `live == ""` → **RED** `digest_unverifiable` (new case, previously fell through to ref comparison)
3. `pinned != ""`, `live != ""`, mismatch → **RED** `digest_mismatch`

The local resolver's empty-digest return is now documented inline with a clear TODO referencing URIRegistry (#167).

**New test:** `TestComputePlan_DeclaredDigest_EmptyLive_Red` in `internal/providers/pin/pin_test.go`

---

### Bug 3 — `cog pin verify` ignores plan's digest/unreachable state

**Codex finding:** `cmd_pin.go:206-238` computes the plan then discards it, re-resolving each pin's live ref with a second inline call. The second pass only compared refs — digest mismatches and unreachable annotations were invisible to verify output.

**Fix commit:** `d6bd26d` — `fix(cmd/pin): verify reads plan actions for digest/unreachable state`

`cmdPinVerify` now consumes `plan.Actions` directly, keyed on the `"reason"` detail: `in_sync`, `digest_unverifiable`, `digest_mismatch`, `target_unreachable`. The output classification is now structurally identical to what `Health()` reads. The now-unused `resolveOneLive` / `localPinResolver` helpers were removed.

**New tests:** `TestComputePlan_VerifyDigestMismatch_PlanHasReason` and `TestComputePlan_VerifyUnreachable_PlanHasReason` in `internal/providers/pin/pin_test.go`

---

### Test commit

`5dae97a` — `test(pin,daemon): add three tests for Codex review findings`

All tests pass: `go test ./internal/providers/... ./pkg/reconcile/...` green. Build clean.

chazmaniandinkle:
## Codex re-review follow-up: three new fix commits

Three commits address the Codex re-review findings on this PR. Force-pushed to `cogos-165-pin-reconcilable`.

---

### Bug A — Health probe mutates state via ApplyPlan

**Codex finding:** `pinProvider.Health()` called `impl.Reconcile()`, which runs the full cycle including `ApplyPlan`. For `sync: read-write` pins this writes pin YAML via `bumpPinFile` — a health probe must never mutate state.

**Fix commit:** `56a1862` — `fix(pin,daemon): health probe is read-only; no ApplyPlan from Health()`

Added `Provider.RefreshState(ctx, root)` to `internal/providers/pin/pin.go` — runs `LoadConfig → FetchLive → ComputePlan` only, no `ApplyPlan`, no filesystem writes. `daemon.pinProvider.Health()` now calls `RefreshState` instead of `Reconcile`.

Approach: Approach 1 (add `RefreshState` method) rather than having the daemon call each phase separately. Chosen because it keeps the read-only cycle in the pin package where the ADR knowledge lives, leaves daemon.go thin, and mirrors how the full `Reconcile` method bundles its phases.

---

### Bug B — Concurrency hole: parallel stale refreshes

**Codex finding:** `pinProvider.Health()` released `p.mu` before calling `Reconcile` and only updated `lastProbe` after success. Two concurrent stale callers could both pass the staleness gate and run parallel `Reconcile` / `RefreshState` calls, racing on `pinStates` writes.

**Fix commit:** `b333e7b` — `fix(daemon): serialise concurrent Health() calls via held mutex`

`p.mu` is now held for the entire probe — staleness check, `RefreshState` call, and `lastProbe` update. Concurrent callers block on `Lock()` and re-check staleness on acquire; by then `lastProbe` is fresh and they skip the refresh.

New test: `TestPinProvider_ConcurrentHealth_NoParallelRefresh` — fires 8 goroutines simultaneously; passes `go test -race` clean.

---

### Bug C — No direct `cmdPinVerify` test

**Codex finding:** Tests asserted `plan.Actions` reasons but did not invoke `cmdPinVerify` directly. Reverting the CLI to its old broken form would not have been caught.

**Fix commit:** `3e23e1e` — `test(pin): direct cmdPinVerify tests via pin.RunVerify (Bug C)`

Extracted core verify logic from `cmdPinVerify` into `pin.RunVerify(ctx, root, filterTarget, resolver, w)` — accepts an injectable `GitHeadResolver` and `io.Writer` for testability. `cmdPinVerify` becomes a 5-line shim. Three new tests invoke `pin.RunVerify` directly:

- `TestRunVerify_DigestUnverifiable` — pin declares digest, resolver returns no digest
- `TestRunVerify_DigestMismatch` — declared digest does not match live digest
- `TestRunVerify_TargetUnreachable` — resolver returns `os.ErrNotExist`

All three assert `[DRIFT]` output content and non-nil error. `go test ./...` passes.

PR #356, opened 2026-06-01
fix(engine): MCP tooling polish — URI hint, type-aware validator, FTS quoting

chazmaniandinkle:
Three small fixes to improve MCP tooling robustness.

## Fix 1: URI error hint (#3)

When a cross-workspace `cog://workspace/...` URI is used before the registry is implemented, the error now includes an actionable hint guiding users to the bare `cog:mem/<path>` form.

**Before:** `unknown workspace authority: workspace "foo" in URI "cog://foo/mem/bar"`
**After:** `unknown workspace authority: workspace "foo" in URI "cog://foo/mem/bar" — cross-workspace registry not implemented; for a local doc use the bare form: cog:mem/<path>`

## Fix 2: Type-aware schema validator (#4b)

`missingSchemaIssues` previously flagged every cogdoc missing a `description` field — including types like `session`, `observation-index`, `audit`, and `pointer` that intentionally don't use it.

Added a `descriptionExemptTypes` allowlist and a `frontmatterTypeValue` helper to read the type field before deciding whether to emit `missing_description`.

## Fix 3: Single-word FTS quoting (#6)

`buildFTSQuery` was passing single-word queries through unquoted, making queries like `-foo` or `type:bar` produce FTS5 syntax errors. Now all terms (single or multi-word) are double-quoted before being OR-joined, consistent with multi-word behavior.

---

**Verified:** `go build ./...` and `go test ./internal/engine/... -tags fts5 -count=1` both pass.


chazmaniandinkle:
Superseded by #357 — rebased cleanly onto upstream/main.

PR #370, opened 2026-06-10
feat(conversations): query-aware URI resolver (ratified RFC, R1-R6)

chazmaniandinkle:
## Summary

Implements the [ratified RFC: Query-Aware Conversation URIs](https://github.com/myrgic/cogos/blob/feat/conversations-uri-resolver/.cog/architecture/rfcs/RFC-query-aware-conversation-uris.cog.md) per decisions R1-R6 (ratified 2026-06-10).

RFC path: `.cog/architecture/rfcs/RFC-query-aware-conversation-uris.cog.md`

### Core: URI parser + resolver (`internal/conversations/uri_resolver.go`)

`ParseConversationURI` validates `cog:conversations[/<source>[/<session>]]` URIs with the full query vocabulary:

- `q=` — term-AND; double-quoted phrase = exact substring (reuses existing `parseSearchQuery`)
- `role=` — comma-list; user/assistant/tool/system
- `since=`/`until=` — RFC3339; `until=` gates bounded-slice hash
- `limit=` (default 20, server-capped at 200) / `offset=` / `order=` (asc|desc)
- `res=` — pointer|abstract|full (ADR-066 resolution ladder)
- `fields=` — field projection
- `component=` and `ontology=`: explicit "reserved, not yet supported" error (R3)
- ALL unknown params: hard error — never silently ignored (R6)

Fragments: `#id-<uuid>` (canonical), `#turn-N`, `#turn-N..M` (positional). Every resolved turn carries its `#id-` anchor for durable re-referencing (RFC §4).

Bounded-slice rule (R5): `content_hash` computed only when `until=` is set or a fragment fully bounds the slice. Unbounded slices → `bounded: false`, no hash.

Response metadata: `{uri, resolved_at, count, sources, bounded, content_hash?}` (RFC §5.3 envelope shape).

### MCP surface (`uri` param on two existing index-backed tools)

`cog_search_conversations` and `cog_get_conversation_turn` now accept `uri=` as an alternative to their existing filter params. Mixing `uri` with other params → explicit error. `cog_list_conversations` is unchanged (index-listing, not turn-addressing).

**Manifest-honesty fix**: all three tools now registered via `MCPServer.TrackTool` (new exported wrapper around `trackTool`). Previously, conversations tools bypassed `trackTool` and were invisible in `/v1/manifest` — this closes the audit gap from 2026-06-10.

**Topology note**: `cog_read_conversation` (sidecar-backed JSONL reader) is NOT touched. It operates on raw filesystem paths, not the index, so URI addressing via the index resolver does not apply.

### HTTP surface (`GET /v1/uri/resolve?uri=cog:conversations/…`)

New route registered alongside the existing `/v1/resolve`. Routes `cog:conversations` URIs to the wired `ConversationsResolver`; all other `cog:` URIs fall through to the existing filesystem path-lookup behaviour.

### Wiring

- `z_conversations_uri_resolver.go` (new, root package): adapts `Provider.ResolveURI` to `engine.ConversationsResolver` interface; registered at boot via `engine.SetConversationsResolver`.
- `z_conversations_mcp.go` + `cmd/cogos/z_conversations_wire.go`: updated to pass `srv.TrackTool`.

## R1-R6 compliance

| Decision | Requirement | Status |
|----------|-------------|--------|
| R1 | Path forms `cog:conversations[/<source>[/<session>]]`; cross-workspace authority → `ErrUnknownAuthority` | DONE |
| R2 | Query vocabulary q/role/since/until/limit/offset/order/res/fields | DONE |
| R3 | `component=`/`ontology=` reserved rejection; unknown params error | DONE |
| R4 | Fragments `#id-`/`#turn-N`/`#turn-N..M`; canonical `#id-` anchor on every turn | DONE |
| R5 | Bounded-slice hash rule (refuse unbounded; hash bounded) | DONE |
| R6 | Read-only; ground-truth refs on turns; schema-at-boundary validation | DONE |

## Test coverage (`internal/conversations/uri_resolver_test.go`)

- Path form parsing (whole observatory, source-scoped, session-scoped)
- Every query param: q, role, since, until, limit, offset, order, res, fields
- Reserved param rejection: `component=`, `ontology=` (with "reserved" in error)
- Unknown param rejection (ref=, format=, cursor=, foo=)
- Phrase vs term: `q="harness attestation"` → single phrase term
- Fragment forms: `#id-`, `#turn-N`, `#turn-N..M`, reversed range error, empty id error
- Bounded hash: no hash on unbounded slice; `sha256:` prefix on `until=`; hash on fragment-bounded
- URI-mixing error (uri + other filter params)
- res= levels: pointer (no text), abstract (truncated), full (complete)
- IDAnchor `#id-<uuid>` on every resolved turn
- Limit + offset arithmetic
- Role filter correctness

`go test -short -timeout 120s ./...` — all packages pass.


chazmaniandinkle:
## e2e wiring fix (pushed: 2c05f12 + c4d2967)

Real-data review found `GET /v1/uri/resolve` answering "conversations resolver not wired" on a live kernel. Two bugs fixed:

**1. Dead wiring (the repo-trap).** `engine.SetConversationsResolver` was called from the repo-root `package main` (legacy cog CLI monolith) — never linked into the `cmd/cogos` daemon. Moved the adapter into `cmd/cogos/z_conversations_wire.go` per the existing hook idiom; deleted the root-package file. Added a binary-assembly regression guard (`cmd/cogos/z_conversations_wire_test.go`) asserting the engine seam is populated after the daemon's init() chain and that param-validation errors propagate through it — the test class that was missing when all unit tests passed against a broken daemon.

**2. MCP uri-only calls were schema-rejected.** The go-sdk marks non-`omitempty` struct fields as required in the generated input schema, so `cog_search_conversations{uri:...}` failed validation ("missing properties: [query]") before the handler ran. `query`/`session_id`/`turn_index` are now `omitempty`; the handlers keep enforcing the runtime contract.

## Live e2e results (sandbox kernel, real hermes ingest data, 1,051 sessions)

HTTP `/v1/uri/resolve`:
- `cog:conversations/hermes-cog/20260525_130925_cf6552?until=2026-06-01T00:00:00Z&res=full` → count=1, bounded=true, content_hash=sha256:138d5b07…
- `cog:conversations?q=BlockedKVCache worktree&limit=3` → count=3, bounded=false, NO content_hash
- `cog:conversations/hermes-cog/20260525_130925_cf6552#id-hermes-cog:1` → exactly 1 turn, text starts `"Say 'Cog online via Eclipse 26B'"`
- `component=` → reserved error; `frobnicate=` → unknown-param error

MCP `POST /mcp` (tools/call):
- `cog_search_conversations{uri:…until…}` → count=1, bounded=true, hash identical to HTTP path (sha256:138d5b07…)
- `cog_get_conversation_turn{uri:…#id-hermes-cog:1}` → count=1, same text
- uri+query mixing → isError=true, "uri param must not be combined with other filter params"
- legacy query-param call → still works (count=2)

**Out-of-scope observation** (pre-existing, not introduced here): the daemon's two reconcile drivers (reconcile-daemon vs autonomic self-heal) resolve different workspace roots for the conversations provider and rewrite the index against each other every ~30s cycle (`deletes=89` / `actions=1140` alternating in the kernel log). This PR touches no reconcile code; flagging for a separate issue.

PR #404, opened 2026-06-24
fix(worktree): stop acknowledged alarms pinning Degraded; remove LoadConfig init race

chazmaniandinkle:
**REVIEW-REQUIRED — left open for your sign-off** (touches worktree reconcile health/init semantics; the audit asked to review the concurrency cluster together).

Two related worktree-reconciler defects from the 2026-06-25 deep audit.

## C2 (HIGH, `worktree_reconciler.go:648`) — perpetual Degraded after acknowledgement
`ApplyPlan`'s health loop counted **any** action whose `Details["classification"]` is an alarm type — including the `ActionSkip` that `ComputePlan` emits for an already-acknowledged alarm (it dedupes the ledger emit but leaves the classification in Details). So once any worktree alarm fired, the provider stayed `HealthDegraded` on **every** subsequent cycle, which kept the autonomic ticker's `needsHeal` true and re-ran a full heal cycle for this provider every tick. That continuous heal is the **steady-state trigger** for the daemon-vs-ticker concurrent-apply hazard (C1).

**Fix:** skip `ActionSkip` actions in the health check — only actively-firing (`ActionUpdate`) alarms degrade health.

## C3 (`worktree_reconciler.go:238`) — data race on lazy adapter init
`LoadConfig` did a nil-check-then-assign on `LedgerReader`/`LedgerWriter`/`GitAdapter`. The daemon and the autonomic ticker call `LoadConfig` on the same instance concurrently, racing those interface-field writes (torn-read risk → possible nil-deref panic in `FetchLive`).

**Fix:** default the adapters in `NewWorktreeReconciler` so the fields are never written at runtime; remove the lazy init. Production registers with nil adapters, so this is the live path. Also removes a latent bug where the ticker's `LoadConfig("")` could mis-init adapters with a cwd-relative root if it ran first.

## Interaction with C1 (not in this PR)
C1 (`autonomic_ticker.go:344` — route `healDegradedProviders` through the daemon's `Trigger` so there's one apply path) is the **root** concurrency fix and is left for a separate, carefully-reviewed change. C2 and C3 are independent strict improvements: C2 removes the continuous trigger; C3 closes the field race. Neither conflicts with a later C1. Per the audit's caution, I did **not** assume C2+C3 fully resolve the concurrency — C1 is still needed for the duplicate-apply-path elimination.

## Tests
The alarm-idempotent test now asserts `Degraded` while the alarm fires, then `Healthy` after acknowledgement (the C2 regression — fails without the guard). New constructor test asserts nil adapters are defaulted (C3). `-race` suite green.

Audit ref: `AUDIT-2026-06-25-deep.md` (C2/C3).

### DIFF
--- a/internal/engine/worktree_reconciler.go
+++ b/internal/engine/worktree_reconciler.go
@@ -200,6 +200,23 @@ type WorktreeReconciler struct {
 // via init(); the reconciler will use defaults wired in LoadConfig once the
 // workspace root is known.
 func NewWorktreeReconciler(repoRoot string, reader LedgerReader, writer LedgerWriter, git GitAdapter) *WorktreeReconciler {
+	// Default the adapters here rather than lazily in LoadConfig. The reconcile
+	// daemon and the autonomic ticker both call LoadConfig on the same instance
+	// concurrently; the lazy nil-check-then-assign there was an unsynchronized
+	// write to these interface fields (a data race, torn-read risk). Doing it
+	// once at construction means the fields are never written at runtime, so all
+	// later reads (including FetchLive's GitAdapter use) are race-free.
+	// Production registers with nil adapters (internal/providers/all), so this is
+	// the live path, not just a convenience for callers that pass their own.
+	if reader == nil {
+		reader = NewFilesystemLedgerReader(repoRoot)
+	}
+	if writer == nil {
+		writer = NewFilesystemLedgerWriter(repoRoot)
+	}
+	if git == nil {
+		git = NewCLIGitAdapter()
+	}
 	return &WorktreeReconciler{
 		RepoRoot:     repoRoot,
 		LedgerReader: reader,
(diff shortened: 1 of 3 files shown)


chazmaniandinkle:
Adversarial review verdict: **MERGE**. Verified: (C2) ComputePlan only emits ActionSkip for *already-acknowledged* alarms — no actively-firing alarm maps to ActionSkip, so the health guard can't hide a real alarm; error reporting (ApplyFailed) is a separate path and untouched. (C3) `repoRoot == cfg.WorkspaceRoot` for the sole production caller (providers/all → RegisterWorktreeReconciler) and the daemon's LoadConfig root (both = cfg.WorkspaceRoot via boot); no caller depends on nil adapters; race-freedom holds (set once before UpsertProvider publication). `-race` green. Rebased onto current main (was cut pre-#403). **Left open for your sign-off** since it's the worktree concurrency cluster; C1 (autonomic_ticker:344 root fix) is intentionally deferred.

PR #408, opened 2026-06-24
perf(reconcile): carry rendered projection in ApplyPlan instead of re-reading the corpus

chazmaniandinkle:
**REVIEW-REQUIRED — left open for your sign-off** (reconcile `ApplyPlan` data flow). Audit Q1 + Q2.

`ProjectionReconciler.ApplyPlan` (`projection_reconciler.go:387`) and `DecisionLineageReconciler.ApplyPlan` (`decision_lineage_reconciler.go:219`) both re-read + re-parse the **entire** nodes/decision corpus and re-render the projection — work `FetchLive` + `ComputePlan` already did earlier in the same reconcile cycle. The `Reconcilable.ApplyPlan(ctx, plan)` signature carries no live state, so the content was re-derived "for idempotency."

But the rendered string is deterministic given (corpus, clock tick), so carrying it forward preserves idempotency **without** the redundant O(corpus) disk scan + YAML parse. With 6 projection instances, a drifted cycle did up to **12 full corpus reads**; the `ProjectionWatcher` fsnotify also fires it on every node write.

## Fix
`ComputePlan` stores the already-rendered projection in `Action.Details["content"]`; `ApplyPlan` writes that directly. A defensive fallback re-derives **only** when `content` is absent (a hand-built / test plan). The atomic tmp+rename and the written bytes are unchanged; the per-action logs use the count already in `Details`.

## Scope / left out
Q3 (`site.go:324`, the same pattern for site CRDs) is **not** included — it touches the site **deploy** path (writes CNAME etc.), which is more consequential than writing an idempotent projection `.md`. Documented in the audit report for a separate, careful change.

## Test
`TestDecisionLineageApplyPlan_UsesDetailsContent`: ApplyPlan writes a sentinel carried in `Details` verbatim (proving no corpus re-derive). Existing full-cycle reconciler tests green under `-race`.

Same per-cycle-rescan elimination shape as #399. Audit ref: `AUDIT-2026-06-25-deep.md` (Q1/Q2).

### DIFF
--- a/internal/engine/decision_lineage_reconciler.go
+++ b/internal/engine/decision_lineage_reconciler.go
@@ -163,6 +163,9 @@ func (r *DecisionLineageReconciler) ComputePlan(config any, live any, state *rec
 				"path":           projPath,
 				"decision_count": len(decisions),
 				"kind":           string(ProjectionDecisionLineage),
+				// Carry the already-rendered projection so ApplyPlan writes it
+				// directly instead of re-loading + re-rendering the whole corpus.
+				"content": projected,
 			},
 		})
 		if action == reconcile.ActionCreate {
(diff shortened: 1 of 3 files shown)


chazmaniandinkle:
Adversarial review verdict: **MERGE**. Verified: (a) Details[content] is byte-identical to the old apply-time re-derive — both renders are deterministic over pre-sorted slices (`TestComputeManifold_Deterministic` proves it; the one map is iterated in fixed key order), and the ONLY divergence is the `_Last generated:` timestamp line, which `equalIgnoringTimestamp` strips, so no churn/idempotency break. (b) Fallback path byte-unchanged. (c) No plan serialization between ComputePlan→ApplyPlan in any production driver, so content stays string + counts stay int. (d) No log/state bloat — content lives only in the in-memory plan for one cycle; BuildState/WriteState derive from live, not Details. (e) Both Create/Update branches carry content; only ActionSkip doesn't and it's short-circuited. Nit (optional): a `_, ok := Details[content]` presence check would be marginally more robust than `== ''`, but both renderers always emit non-empty content so it's moot. Q3 (site deploy path) correctly deferred.

Issue #441, opened 2026-07-06
perf(inference): make OpenAI-compat /v1/models discovery lazy instead of polling every 10s


chazmaniandinkle:
## Summary

The kernel's background availability maintainer probes every registered OpenAI-compatible provider on a fixed **10-second ticker**, and each probe issues a live `GET /v1/models`. For a local LM Studio provider (`lmstudio-darkstar` at `127.0.0.1:1234`) this produces a continual, unsolicited `GET /v1/models` ping roughly every 10s for the life of the process, cluttering the LM Studio request log even when no inference is happening. The desired behavior is **lazy / event-driven** model discovery: probe on demand (first use / cache miss) and/or on an explicit invalidation signal, not on an unconditional wall-clock timer.

The fix should cover the **local** provider path specifically. A separate Eclipse LAN node (`192.168.10.191:1234`) already has a 30s TTL debounce in-tree (see Root cause), so this is about bringing the local path up to (or past) that bar, not re-inventing the pattern.

## Symptom

LM Studio's request log at `127.0.0.1:1234` shows a steady stream of `GET /v1/models` requests at a fixed cadence ("every few seconds") with no user-initiated inference in flight. The cadence is regular, not bursty, and continues as long as the kernel is running.

## Root cause

Verified against current canonical `internal/engine` source (line numbers as of the dev tree; the observed behavior is on the running **v0.16.15** release binary — the mechanism is identical). There are two categories of caller: one eager one-shot probe (benign) and one **unconditional periodic loop** (the driver of the observed pings). They must not be conflated.

**Primary driver — background 10s availability ticker (unconditional, no client needed):**

- `internal/engine/router.go:48` — `const defaultAvailTTL = 10 * time.Second`
- `internal/engine/router.go:320` — `SimpleRouter.Start(ctx)` spawns the maintainer goroutine
- `internal/engine/router.go:322` — `availTTL` defaults to `defaultAvailTTL` (10s) when unset
- `internal/engine/router.go:326` — `t := time.NewTicker(r.availTTL)`
- `internal/engine/router.go:334-335` — on each `<-t.C` tick, calls `r.probeAll(ctx)`
- `internal/engine/router.go:301-312` — `probeAll` fans out one goroutine per provider, each calling `p.Available(pctx)` (line 307) under `probeTimeout`
- `internal/engine/provider_openai.go:104-105` — `OpenAICompatProvider.Available()` calls `listModels(ctx)`
[...]

PR #443, opened 2026-07-06
perf(inference): cache OpenAICompatProvider.Available() to stop 10s /v1/models polling (#441)

chazmaniandinkle:
## What

Adds a per-provider 30s TTL cache to the live `Available()` reachability probe (`GET /v1/models`) for both `OpenAICompatProvider` and `ClaudeOAuthProvider`, so the router's availability ticker and the per-request `/v1/providers` handler stop hammering the upstream on every call.

## Why — refs #441

The router's background availability maintainer probes every registered provider on a fixed **10-second ticker** (`defaultAvailTTL`, `internal/engine/router.go`), calling `Available()` → `GET /v1/models`. For an OpenAI-compatible provider (LM Studio, vLLM, …) this floods the server's request log every ~10s even when idle; for `ClaudeOAuthProvider` the same ticker hits the **paid** Anthropic endpoint. The per-request `/v1/providers` handler (`serve_compat.go`) calls `Available()` too, uncached.

## How

- Shared `availCacheTTL` (30s) const beside `probeTimeout` in `router.go`.
- `OpenAICompatProvider.Available()` and `ClaudeOAuthProvider.Available()` return a cached result within the TTL; on expiry they run one probe under a mutex (held across the HTTP call so concurrent callers collapse into a single request). The live logic is factored into each provider's `probeAvailable()`. This mirrors the existing probe cache on `MLXSupervisedProvider`.
- **Poisoning guard:** a negative result caused only by the caller's cancelled/expired context (`ctx.Err() != nil`) is **not** cached — a client disconnecting mid-probe can't mark a healthy provider unavailable to everyone for the TTL.

Drops upstream `/v1/models` probes from ~1/10s to ~1/30s for the whole class of OpenAI-compat providers **and** the OAuth provider, and covers the `/v1/providers` path for free.

## Tests

- `TestOpenAIAvailableCachesWithinTTL`, `TestClaudeOAuthAvailableCachesWithinTTL` — 6 calls → exactly 1 upstream hit.
- `TestOpenAIAvailableCancelledContextDoesNotPoisonCache` — a cancelled probe does not poison the shared cache.
- Existing `Available` tests unchanged and green. `go test ./internal/engine/ -race` passes; `go build ./...` clean.

## Scope / follow-up

**Option A** from #441 — caps cadence at ~1/TTL; it does **not** make an idle kernel fully silent (the 10s ticker still refreshes on expiry). **Option B** (retire the wall-clock ticker for local providers in favor of on-demand priming + `ReconcileDaemon.Trigger` invalidation) remains the follow-up for true idle-silence, leaving #441 open. TTL is a single shared const, trivially tunable / config-drivable.

### DIFF
--- a/internal/engine/provider_claudeoauth.go
+++ b/internal/engine/provider_claudeoauth.go
@@ -604,9 +604,11 @@ func (p *ClaudeOAuthProvider) Available(ctx context.Context) bool {
 		return p.availResult
 	}
 	fresh := p.probeAvailable(ctx)
-	// Don't cache a negative caused only by the caller's cancelled/expired
-	// context (#441 review) — it says nothing about provider health.
-	if !fresh && ctx.Err() != nil {
+	// Skip caching only for a caller-initiated cancellation (client disconnect);
+	// a context deadline is the router's probeTimeout firing on a slow provider
+	// and must be cached as unavailable, else a hung provider stays cached as
+	// available forever (#441 review).
+	if !fresh && ctx.Err() == context.Canceled {
 		return p.availResult
 	}
 	p.availResult = fresh
--- a/internal/engine/provider_claudeoauth.go
+++ b/internal/engine/provider_claudeoauth.go
@@ -533,6 +533,14 @@ type ClaudeOAuthProvider struct {
 	client    *http.Client
 	lc        *CredentialLifecycle
 
+	// availMu guards the Available() TTL cache (#441): the router's 10s
+	// availability ticker probes this provider with a live GET /v1/models against
+	// the (remote, paid) Anthropic endpoint. Cache the result for availCacheTTL,
+	// holding the mutex across the probe so concurrent callers collapse into one.
+	availMu     sync.Mutex
+	availResult bool
+	availAt     time.Time
+
 	// fallback is invoked on a persistent 429 (subscription burst rate-limit
 	// with overage disabled). Typically a claude-code CLI provider that reaches
 	// the same Max subscription via the official client. nil disables fallback.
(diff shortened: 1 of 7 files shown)


chazmaniandinkle:
Thanks — all three findings were valid and are addressed in 48e0afe.

**1. Context poisoning (the real regression).** Fixed. `Available()` now skips the cache write when the probe is negative *and* the caller's context is done (`!fresh && ctx.Err() != nil`), so a client disconnecting on `/v1/providers` (which passes `r.Context()`) can't cache a false negative that marks a healthy provider down for the whole TTL for everyone — including the router's background `probeAll`. Regression test `TestOpenAIAvailableCancelledContextDoesNotPoisonCache` (cancelled ctx → not cached → next healthy caller still sees available).

**2. `ClaudeOAuthProvider` left unfixed.** Fixed — agreed it's the more pressing instance since it probes the paid remote Anthropic endpoint. Gave `ClaudeOAuthProvider.Available()` the same TTL cache + poisoning guard (its live body is now `probeAvailable()`). Test `TestClaudeOAuthAvailableCachesWithinTTL` (6 calls → 1 upstream hit).

**3. Nonexistent `eclipseProbeCache` precedent.** Correct — that symbol lives only in a *downstream consumer copy* of `serve.go`, not this repo; I conflated the two. Removed the citation, hoisted the TTL into a shared `availCacheTTL` const beside `probeTimeout` in `router.go`, and pointed at the real in-repo precedent (`MLXSupervisedProvider`'s probe cache). PR description updated too.

On the unverified note (a genuinely-down provider now caches unavailable for up to 30s, slowing recovery detection from ~10s): that's the intended debounce tradeoff, now bounded by `availCacheTTL`, and the guard ensures only *real* negatives (not cancellations) are cached. Happy to shorten the TTL or make it config-driven if faster recovery is preferred.

`go build ./...`, `go vet ./internal/engine/`, and `go test ./internal/engine/ -race` all green.

chazmaniandinkle:
Round 2 addressed in aad233d — both findings were correct.

**1. Guard conflated cancellation vs. probeTimeout deadline.** Fixed. The guard now skips the cache write **only** for `context.Canceled` (a caller/client disconnect), not `context.DeadlineExceeded`. So a provider that degrades past the router's 2s `probeTimeout` now caches `false` (unavailable) as it should, instead of discarding the negative and keeping stale `true` forever. New regression test `TestOpenAIAvailableCachesDeadlineExceededAsUnavailable`: a hung server + a deadline context → `Available()` returns false, the negative is cached (second call served from cache, exactly 1 upstream hit).

**2. `OllamaProvider.Available()` unfixed.** Fixed — same TTL cache + guard, test `TestOllamaAvailableCachesWithinTTL`. I swept the rest of the class to avoid another round: `AnthropicProvider` (`apiKey != ""`), `Codex`/`ClaudeCode`/`Pi` (`exec.LookPath`), and `Stub` do **no** network I/O in `Available()`, and `MLXSupervisedProvider` already caches its probe — so Ollama was the last live-probe provider.

`go build ./...`, `go vet ./internal/engine/`, and `go test ./internal/engine/ -race` all green.

Issue #442, opened 2026-07-06
refactor(self-update): make kernel.toml the single kernel-version SOT, maintained write-ahead by self-update (kill self-update.yaml/state version fragmentation)


chazmaniandinkle:
## Summary

`kernel.toml` is documented as the single source of truth for the kernel binary version + per-platform checksums, and is meant to be written at install and kept reconciled with the running daemon by self-update. In practice the daemon's self-update path routes AROUND `kernel.toml` entirely and establishes a parallel version authority (`self-update.yaml` policy + an in-process `running_version` + a derived `.state.json`). The result is version-SOT fragmentation: three artifacts that have never been reconciled with each other.

This issue proposes making `kernel.toml` the actual single SOT: install writes it, and self-update MAINTAINS it write-ahead (ledger-first) — update `kernel.toml` first, then download+verify against that checksum, then swap+health-verify, then roll back both binary and `kernel.toml` on failure. `self-update.yaml` is reduced to policy only, carrying no version truth.

> Provenance note: the self-update provider internals cited below live in the `myrgic/cogos` daemon source (`internal/providers/selfupdate/…`) and were read via topology/pin/install surveys, not re-verified against `main` in this pass. One survey notes the provider may exist in a worktree not yet merged to main. Claims about provider internals are marked accordingly and should be confirmed against current `main` before implementation.

## The anti-pattern: fragmented version SOT

Three artifacts each carry (or imply) a kernel version, and they disagree:

| Artifact | Role | Observed value | Status |
|---|---|---|---|
| `.cog/conf/kernel.toml` `[kernel].version` | install-time SOT (shell) | v0.16.15 (workspace hand-corrected this pass; historically v0.8.0) | authoritative-by-doc, but daemon never reads it |
| `.cog/config/self-update.yaml` | daemon runtime policy | `enabled: true, channel: stable, auto_apply: true` | LIVE authority — carried the daemon forward |
| in-process `running_version` (via `SetRunningVersion`) | daemon memory | the actual running tag | authoritative-in-fact, unpersisted to any SOT |
| `.cog/config/self-update/.state.json` | derived reconcile state | `running_version v0.16.11, target_tag v0.16.12, in_progress: true` | STALE relative to reality (daemon is at v0.16.15) |

The daemon-owned policy (`self-update.yaml`) is what actually advanced the kernel; `kernel.toml` was written once at install and then bypassed; `.state.json` is stale. No path reconciles them.

## Current topology (from surveys)

[...]

PR #444, opened 2026-07-06
fix(self-update): harden write-ahead kernel.toml maintainer (Fable review)

chazmaniandinkle:
## What

Introduces the daemon-side write-ahead maintainer for `.cog/conf/kernel.toml`
(the pinned kernel version + per-platform sha256 checksums SOT, issue #442) and
hardens it against the correctness and premise findings from a rigorous Fable
review. Every issue below was reproduced by direct probe before fixing.

New file `internal/engine/cli_selfupdate_kerneltoml.go` records, BEFORE the
binary is downloaded, the target `[kernel].version` + the current platform's
sha256 into `kernel.toml` (the ledger entry the swap reconciles toward), and
restores the prior bytes on any later failure so the recorded version never
claims a binary that is not actually running.

## Why

The self-update apply path previously routed around `kernel.toml`, so any file
that pins the version/checksums would drift stale after every auto-update. The
write-ahead maintainer closes that drift. The review surfaced four ways the
first cut of this file could either corrupt the SOT it is meant to keep correct
or misrepresent its own state; this PR fixes all four.

## How

**Correctness — `patchKernelTOML`**

- **Quoted checksum keys are now matched.** A quoted `"darwin-arm64" = "..."`
  (valid TOML, common for hyphenated keys) did not compare equal to the bare
  platform key, so the current-platform line was not replaced in place and a
  **second** `darwin-arm64 = "..."` line was appended — duplicate-key TOML that
  strict parsers reject, defeating #442 on exactly the file it maintains.
  `tomlLineKey` now normalises a quoted key (double- or single-quoted) to its
  bare form for matching, and `replaceTOMLValue` preserves the on-line key text
  so the entry is rewritten in place with its quoting intact. One entry, updated.
- **Two non-canonical layouts now fail loudly** instead of emitting invalid
  TOML: (a) a bare top-level `version` before any `[kernel]` header (a prepended
  fresh `[kernel].version` would duplicate the key and strand the stale value);
  (b) checksums present only as `[kernel.checksums.<other>]` subtables with the
  current platform absent (appending an inline `[kernel.checksums]` after its
  child subtable is a parent-redefined-after-child error). Erroring aborts the
  write-ahead in the caller with the running binary untouched — strictly safer
  than corrupting the SOT.

**Permission preservation — write-ahead + rollback**

- `atomicWriteConfigFile` writes via `os.CreateTemp` (0o600) + rename and never
  applied the captured `priorPerm`, so a 0o644 `kernel.toml` silently narrowed
  to 0o600 and stayed there even after rollback. `writeAheadKernelTOML` and
  `rollbackKernelTOML` now `os.Chmod` back to the snapshot's `priorPerm`, so the
  file keeps its original mode across both write-ahead and rollback.

**Premise / documentation**

- The original file header asserted a broken-then-fixed state that cannot be
  reproduced against this tree: there is **no** `.cog/conf/kernel.toml`, **no**
  installer that seeds it, and **no** reader of `[kernel.checksums]` anywhere in
  the repo. The header is corrected to state the maintainer is
  **dormant-by-design** — an absent `kernel.toml` is a silent no-op, so the
  binary swap/restart/rollback sequence is byte-for-byte unchanged on every
  current workspace (hence no regression risk to the running daemon) — and only
  becomes active once a separate change lands the SOT file plus the installer
  (which, for an external installer, may live outside this repo).

## Tests

- Quoted- and single-quoted-key in-place replacement: exactly one entry, stale
  value gone, quoting preserved.
- Both malformed layouts assert `patchKernelTOML` returns an error.
- `TestWriteAheadKernelTOML_WritesThenRollsBack` now asserts file mode `0o644`
  is preserved across write-ahead and restored on rollback.
- Existing write-ahead integration tests (`TestRunApplyWriteAhead*`, seeded with
  the canonical inline layout) continue to pass — the new error paths only fire
  on the non-canonical hand-edited layouts.
- `go build ./...`, `go vet ./internal/engine/`, and
  `go test ./internal/engine/ -race -count=1` all pass (go1.25.5).

## Review provenance

Addresses the findings of a Fable review across two lenses (correctness of the
write-ahead ledger; no-regression + conventions for the apply/restart/rollback
path). The two `major`, two `minor` findings are all resolved here.

### DIFF
--- a/internal/engine/cli_selfupdate_kerneltoml.go
+++ b/internal/engine/cli_selfupdate_kerneltoml.go
@@ -1,0 +1,382 @@
+// cli_selfupdate_kerneltoml.go — write-ahead maintenance of .cog/conf/kernel.toml.
+//
+// kernel.toml is intended as the single source of truth for the pinned kernel
+// binary version + per-platform sha256 checksums: an installer seeds it once and
+// the daemon self-update path is expected to keep it reconciled with the running
+// binary so the recorded version never drifts ahead of what is actually running
+// (issue #442). This file provides the daemon side of that contract.
+//
+// DORMANT-BY-DESIGN until the SOT lands. As of this change the repo does NOT yet
+// carry a .cog/conf/kernel.toml, an installer that writes one, or any reader of
+// [kernel.checksums]; those arrive in a separate change (and, for an external
+// installer, may live outside this repo). writeAheadKernelTOML therefore treats
+// an ABSENT kernel.toml as a silent no-op: on every current workspace the swap
+// sequence is byte-for-byte unchanged and there is no maintenance to perform.
+// The maintainer only becomes active once a kernel.toml is present to reconcile.
+//
+// When a kernel.toml IS present, this file gives the darwin updater a
+// write-ahead (ledger-first) maintainer: BEFORE downloading the new binary, the
+// updater records the new [kernel].version + the current platform's sha256 into
+// kernel.toml (the ledger entry the swap then reconciles toward). If any later
+// step fails — checksum mismatch, smoke-test, swap, restart, or health
+// verification — the prior kernel.toml bytes (and file mode) are restored so the
+// recorded version never claims a binary that is not actually running.
+//
+// The edit is deliberately surgical (line-based) rather than a full TOML
+// round-trip: the repo carries no TOML library, and a byte-preserving edit is
+// the only way to keep comments, the release/checksums URL templates, and every
+// OTHER platform's checksum entry intact (issue #442 migration note: "never
+// blank non-current platforms"). Semantics mirror the YAML node-patch approach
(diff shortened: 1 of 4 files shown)


chazmaniandinkle:
Fixed in 4ee0a10. The perm-restore `os.Chmod` in both `writeAheadKernelTOML` and `rollbackKernelTOML` is now best-effort (`_ = os.Chmod(...)`). Since `atomicWriteConfigFile` commits via an all-or-nothing rename, `writeAheadKernelTOML` now returns a non-nil error **only** from paths that run before the atomic commit (patch failure, write failure) — so an error always means `kernel.toml` is unchanged, and the caller's abort-without-rollback is safe by construction. `go build`/`vet`/`go test ./internal/engine/ -race` green.

PR #458, opened 2026-07-07
fix(conversations): atomic + cross-process-locked writes to _meta.json

chazmaniandinkle:
## What

Fixes #449 (conversations index multi-writer race on `_meta.json`), and two
related multi-writer races on the same code path discovered while fixing it:
a matching race on each session's turns file, and a delete-vs-upsert
interleaving across the two.

Also extends the same atomic-write + cross-process-lock fix to three more
confirmed sites of the identical bug class, flagged across two more
cog-review rounds (fifth pass, head `9d0aa2e`; sixth pass, head `c5726e2`):
`pkg/substrate/reconcile/state.go` (the generic reconcile framework backing
every other resource-provider type), and both of `internal/eval/mcp_tools.go`'s
JSON sidecars — the dispatch-trigger trigger file and the baseline-pin file
(the latter initially misjudged as out of scope; corrected in the sixth
round). See "5.", "6.", and "7." below.

## How

**1. `_meta.json` (issue #449's original race).**
`internal/conversations/index.go`'s `writeMetaFileLocked` did a plain
`os.WriteFile` with no cross-process coordination. A CLI-invoked
`cog reconcile conversations` run and the daemon's own reconcile cycle could
each read-modify-write `_meta.json` independently; the second writer's
marshal-of-its-own-in-memory-snapshot silently dropped whatever the first
writer had already committed ("last writer wins").

- **Atomic write**: tmp + rename, same pattern as
  `BusSessionManager.saveRegistry` (`internal/engine/bus_session.go:240`).
- **Cross-process lock** (`metaLockPath`, a sibling `_meta.json.lock` file)
  around the full read-modify-write cycle.
- **Delta application, not snapshot merge**: each write applies only the
  calling method's own delta (one upsert or one delete) on top of a fresh
  read of the on-disk state, never a full re-marshal of the process's
  in-memory `idx.sessions` (which could be stale relative to a peer's more
  recent write and would silently resurrect a session the peer already
  deleted).
- The disk-merged map is only adopted into `idx.sessions` (and the
  `LoadIfChanged` fast-path bookkeeping refreshed) when it's provably
  identical to what this process already has fully loaded (turns included).
  Otherwise the next `LoadIfChanged` detects the on-disk change and performs
  a real reload, keeping `idx.sessions` and `idx.turns` consistent with each
  other (this refinement came out of a cog-review round — see commit
  history for the earlier, unsound version and why).

**2. Each session's turns file (same race, different file).**
`writeTurnsFile` had the identical non-atomic, non-locked shape as the
original `_meta.json` bug, on a path this PR already touches. Fixed the same
way: atomic tmp+rename, plus a per-sessionID cross-process lock
(`turnsLockPath`) so two processes racing on the *same* session's turns file
fully serialize instead of tearing each other's write.

**3. Delete-vs-upsert interleaving across the two locks.**
Taking the turns lock and the meta lock as two *separate*,
independently-released critical sections left a window: instance B's turns
write completes and releases the turns lock; instance A acquires that
now-free lock, removes the turns file, and moves on to the meta lock; B then
writes the session back into `_meta.json`. Net result: `_meta.json` lists a
session whose turns file is gone — a split-brain reachable only via this
interleaving. Fixed by holding **one** per-sessionID lock across the entire
`UpsertSession`/`DeleteSession` body (both the turns operation and the meta
write), making the two calls for the same sessionID fully mutually exclusive
end-to-end. Different sessionIDs still use different lock files and proceed
with no contention.

**4. Lock-path collision edge case.**
`turnsLockPath(sessionID)` used to be `turnsPath(sessionID)+".lock"`; for a
session literally named `_meta`, `turnsPath("_meta")` resolves to the same
file as `metaPath()` (`_meta.json`), which would make `turnsLockPath`
collide with `metaLockPath` and self-deadlock `UpsertSession("_meta", ...)`
against its own nested `writeMetaFileLocked` call. Fixed with a distinct
`.sessions.lock` suffix.

**Also audited** `sdk/thread.go:250` (the other shared-file read-modify-write
path named in ledger item L10). Its meta write is already atomic via
`fs.WriteJSON`/`WriteAtomic`. No lock was added there — `sdk.Kernel` (the
only type that constructs the `threadProjector` doing this write) has zero
callers repo-wide outside the `sdk` package itself, so there is no live
CLI+daemon (or any two-process) interleaving on `threadMetaPath` to race on
today. Documented this finding inline so the same lock pattern is applied if
`sdk.Kernel` is ever wired into a live entrypoint.

**5. `pkg/substrate/reconcile/state.go` — the generic reconcile framework's
state file (cog-review finding, fifth pass).** `LoadState`/`WriteState` back
every registered resource-provider type (site, component, agent, service,
discord, eval, rbac_bindings, identity, openclaw, ...) via the identical
`LoadState → ComputePlan → ApplyPlan → BuildState → WriteState` cycle run by
both `cogos reconcile <type>` (`internal/engine/cli_reconcile.go`) and the
daemon's own periodic reconcile loop (`internal/engine/reconcile_daemon.go`)
against the same un-namespaced `.state.json` — the same architecture that
produced issue #449 in the first place, one layer up. `WriteState` was
already atomic (tmp+rename); the read-modify-write *cycle* around it had no
cross-process coordination, so two concurrent cycles' `WriteState` calls race
last-write-wins and silently discard the other cycle's `Serial`/`Resources`.

Unlike the conversations index fix, `BuildState` fully re-derives the entire
`Resources` slice from live state plus the existing state's `Serial`/`Lineage`
each cycle — it does not apply a delta on top of a fresh disk read — so there
is no sound delta-merge available here (a snapshot-merge would hit the same
unsoundness the conversations fix explicitly rejected). The fix is a new
`AcquireStateLock`/`Release` pair (`pkg/substrate/reconcile/state.go`,
`StateLockPath` = a sibling `.state.json.lock` file, `StateLockTimeout` =
5s, matching `metaLockTimeout`) that callers wrap around their entire
`LoadState → ... → WriteState` cycle, wired into both
`cli_reconcile.go:runReconcileCmd` and
`reconcile_daemon.go:runOneCycle` (held via `defer` across every early-return
path — `ComputePlan`/`ApplyPlan`/`BuildState` failures all still release it).

**6. `internal/eval/mcp_tools.go` — the eval dispatch-trigger sidecar (same
cog-review finding).** `writeDispatchTrigger` (called from the
`cog_run_experiment` MCP tool handler) and `readAndClearDispatchTriggers`
(called by the daemon's own eval `ComputePlan` cycle) both did a plain
`os.WriteFile` on `eval-dispatch-triggers.json` — no atomic tmp+rename, no
lock at all, the least-protected instance of this bug class in the repo. A
user calling `cog_run_experiment` at the same moment the daemon is draining
triggers could have the just-added trigger silently wiped by the concurrent
clear (or the clear itself clobbered back to non-empty) with no error
surfaced to the caller — an experiment dispatch silently dropped. Fixed with
the same atomic tmp+rename write (`writeFileAtomic`) plus a cross-process
`filelock` (`dispatchTriggersLockPath`, `dispatchTriggersLockTimeout` = 5s)
held across the full read-modify-write cycle in both functions.

**7. `internal/eval/mcp_tools.go` — the baseline-pin sidecar (sixth
cog-review pass; corrects an earlier judgment call in this same PR).**
`writePinBaseline` (`eval-baselines.json`, same file as the dispatch triggers)
shares the identical unlocked, non-atomic `os.WriteFile` read-modify-write
shape. An earlier round of this PR judged it out of scope on the claim that
no concurrent reader/writer exists — that claim was **incorrect**:
`internal/eval/provider_impl.go`'s `LoadConfig` reads `eval-baselines.json` on
every reconcile cycle (both the daemon's loop and `cogos reconcile eval`),
and two concurrent `cog_pin_baseline` MCP calls race each other with zero
serialization. Fixed the same way: atomic tmp+rename (`writeFileAtomic`,
shared with the dispatch-trigger fix) plus a cross-process `filelock`
(`pinBaselineLockPath`) spanning the read-modify-write cycle.
`LoadConfig`'s read itself needed no lock — once the write is atomic, a
racing plain read can only observe the file whole, pre- or post-rename,
never torn.

## Test evidence

Six new tests in `internal/conversations/index_multiwriter_test.go`, all run
under `-race`:

- `TestCrossProcessUpsertRace` — two `*Index` instances upsert 100 disjoint
  sessions each concurrently; asserts all 200 survive in valid `_meta.json`.
- `TestCrossProcessDeleteSurvivesConcurrentUpsert` — a delete on one instance
  is never resurrected by a concurrent peer upsert of unrelated sessions.
- `TestPeerWriteDoesNotDesyncSessionsAndTurns` — a peer's on-disk addition
  never enters `idx.sessions` without `idx.turns` also covering it; a
  subsequent `LoadIfChanged` correctly backfills both.
- `TestCrossProcessSameSessionTurnsFileRace` — 8 concurrent writers hammering
  the *same* sessionID's turns file 30x each; the file is always valid,
  untorn JSON afterward.
- `TestCrossProcessDeleteUpsertSameSessionNoSplitBrain` — 50 rounds of
  concurrent delete-vs-upsert on the identical sessionID; `_meta.json` and
  the turns file are always consistent (both present or both absent).
- `TestTurnsLockPathDoesNotCollideWithMetaLockPath` — asserts the two lock
  paths differ and that `UpsertSession("_meta", ...)` completes promptly
  rather than hanging on a self-deadlock.

```
$ go test ./internal/conversations/... -race -count=3
ok  	github.com/myrgic/cogos/internal/conversations	21.9s
```

Full repo build: `go build ./...` clean. All pre-existing
`internal/conversations` tests (including the `load_guard_test.go`
sole-writer/external-write/same-size-rewrite trio, which the sole-writer
fast-path change had to keep passing) still pass under `-race`.

**New for the two sibling-completion sites (5. and 6. above):**

`pkg/substrate/reconcile/state_lock_test.go`:
- `TestAcquireStateLock_SerializesConcurrentCycles` — 50 goroutines each
  simulating a full reconcile cycle (`AcquireStateLock` → `LoadState` →
  append one resource → `WriteState` → `Release`); asserts final `Serial` ==
  50 and all 50 `Resources` entries survive (no cycle's write silently lost).
- `TestAcquireStateLock_ContendsOnSameResourceType` — a peer holding
  `StateLockPath` blocks a concurrent `AcquireStateLock` for the same
  `resourceType` (probed under a budget well below `StateLockTimeout`), while
  a different `resourceType`'s lock is unaffected.
- `TestStateLockPath_IsSiblingOfStateFile` — lock path is a sibling file, not
  the state file itself.

`internal/eval/eval_sidecar_multiwriter_test.go` (renamed from
`dispatch_triggers_multiwriter_test.go` once it grew to cover both sidecars):
- `TestDispatchTriggers_ConcurrentWritesAllSurvive` — 50 goroutines each
  calling `writeDispatchTrigger` for a distinct experiment ID concurrently;
  asserts all 50 survive a single `readAndClearDispatchTriggers` call.
- `TestDispatchTriggers_WriteDuringClearDoesNotLoseEitherSide` — races one
  `writeDispatchTrigger` call against a concurrent
  `readAndClearDispatchTriggers` call for a pre-existing trigger; asserts the
  clear always observes the pre-existing entry and the pre-existing entry is
  never resurrected after the clear.
- `TestWritePinBaseline_ConcurrentWritesAllSurvive` — 50 goroutines each
  calling `writePinBaseline` for a distinct experiment ID concurrently;
  asserts all 50 pins survive on disk in valid JSON.

```
$ go test ./pkg/substrate/reconcile/... -race -count=3
ok  	github.com/myrgic/cogos/pkg/substrate/reconcile	9.055s
$ go test ./internal/eval/... -race -count=3
ok  	github.com/myrgic/cogos/internal/eval	8.580s
$ go build ./... && go test ./... -count=1
ok (all packages, full repo)
```

## Acceptance

- [x] Fixes #449
- [x] `go build ./...` clean
- [x] `go test ./internal/conversations/... -race` green (existing + 6 new)
- [x] `sdk/thread.go` sibling audited and disposition documented inline
- [x] Delete/upsert split-brain and lock-collision findings from review
      addressed with regression tests
- [x] `pkg/substrate/reconcile/state.go` sibling race fixed
      (`AcquireStateLock`/`Release` wraps the full cycle in both
      `cli_reconcile.go` and `reconcile_daemon.go`), with regression tests
- [x] `internal/eval/mcp_tools.go:344` sibling race fixed (atomic write +
      cross-process lock on `eval-dispatch-triggers.json`), with regression
      tests
- [x] `writePinBaseline` / `eval-baselines.json` sibling race fixed (same
      atomic write + cross-process lock), with regression tests — corrects
      this PR's own earlier out-of-scope judgment call after cog-review
      showed `provider_impl.go`'s `LoadConfig` does read this file every
      reconcile cycle

---

**Update (round 4, head fix/shared-file-atomic-writes):** cog-review flagged the root-package `busSessionManager.saveRegistry` (bus_session.go:224) as an un-swept sibling. Addressed: made it atomic (tmp+rename), closing the truncation→total-registry-loss failure mode (the severe half). The remaining cross-process last-writer-wins race between this root-CLI writer and the daemon writer of the same registry.json requires a shared filelock on BOTH sides; the daemon half is engine bus-registration behavior and the root half is L1-dispositioned, so full cross-process coordination is deferred to #461 (not silently dropped). This PR closes #449 and the atomic-write half of the whole class; #461 tracks the cross-process-lock remainder.

### Registry.json cross-process lock (0b33810)
Both writers of `.cog/.state/buses/registry.json` (root CLI `busSessionManager` and daemon `internal/engine.BusSessionManager`) now take the same `registry.json.lock` advisory filelock around every load-modify-save cycle, closing the last-writer-wins entry-drop the review's round-N finding described. Regression test: `TestRegisterBus_ConcurrentManagersAllSurvive` (independent manager instances = separate-process simulation). Nothing on this file is deferred to L1.

### Lock-ordering fix (082ecb7)
Round-N+1 review correctly flagged that the new registry filelock was acquired inside `m.mu` on the append hot path (nested-lock class this PR fixes elsewhere). Fixed: seq updates run outside `m.mu` with a 500ms best-effort timeout and a monotonic `IfNewer` guard (out-of-order updates are no-ops, not regressions); registration orders filelock BEFORE `m.mu`; engine side gains the multiwriter + monotonic regression tests it was missing. Blast radius: append latency under peer contention is now bounded by 500ms best-effort skip instead of 5s in-mutex stall; no unrelated bus operation queues behind a contended peer.

### DIFF
--- a/bus_session.go
+++ b/bus_session.go
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 	"sync"
 	"time"
+
+	"github.com/myrgic/cogos/pkg/filelock"
 )
 
 // ADR-084 dataref migration policy (Phase 1: schema-additive).
@@ -177,8 +179,31 @@ func (m *busSessionManager) createMCPBus(sessionID, origin string) (string, erro
 	return busID, nil
 }
 
+// registryLockTimeout bounds how long a registry writer waits for the
+// cross-process advisory lock before failing the operation.
+const registryLockTimeout = 5 * time.Second
+
+// acquireRegistryLock takes the cross-process advisory lock guarding the
+// load-modify-save cycle on registry.json. The daemon's
+// internal/engine.BusSessionManager takes the SAME lock (registry.json.lock),
+// so concurrent `cog bus`/`cog infer` invocations and a running daemon
+// serialize instead of last-writer-wins silently dropping each other's
+// registry entries.
+func (m *busSessionManager) acquireRegistryLock() (*filelock.FileLock, error) {
+	if err := os.MkdirAll(m.busesDir(), 0755); err != nil {
+		return nil, err
+	}
+	return filelock.Acquire(m.registryPath()+".lock", registryLockTimeout)
+}
+
 // registerBus adds or updates a bus entry in the registry.
 func (m *busSessionManager) registerBus(busID, sessionID, origin string) error {
+	lock, err := m.acquireRegistryLock()
+	if err != nil {
+		return fmt.Errorf("acquire registry lock: %w", err)
+	}
+	defer lock.Release()
+
 	registry := m.loadRegistry()
 
 	// Check if bus already exists
@@ -229,12 +254,10 @@ func (m *busSessionManager) loadRegistry() []busRegistryEntry {
 // one being added. Mirrors the daemon-side BusSessionManager.saveRegistry
 // (internal/engine/bus_session.go).
 //
-// NOTE: this closes the truncation→total-loss failure mode but does NOT add a
-// cross-process advisory lock. Full serialization of this root-CLI writer
-// against the daemon's writer of the same registry.json requires both sides to
-// take a shared filelock; that coordination is deferred to the L1 disposition
-// of the root bus-CLI (whether this cross-process scenario survives at all
-// depends on relocate-vs-delete). Tracked separately; see PR body.
+// Cross-process serialization: every load-modify-save cycle on the registry
+// (registerBus, updateRegistrySeq) holds the shared advisory lock from
+// acquireRegistryLock; the daemon-side writer holds the same lock. Callers of
+// saveRegistry must hold that lock.
 func (m *busSessionManager) saveRegistry(entries []busRegistryEntry) error {
 	if err := os.MkdirAll(m.busesDir(), 0755); err != nil {
 		return err
@@ -432,6 +455,15 @@ func (m *busSessionManager) getLastEvent(busID string) (int, string) {
 
 // updateRegistrySeq updates the last event seq/timestamp in the registry.
 func (m *busSessionManager) updateRegistrySeq(busID string, seq int, ts string) {
+	lock, err := m.acquireRegistryLock()
+	if err != nil {
+		// Best-effort metadata update: skipping is safer than an unserialized
+		// read-modify-write that can drop a concurrent writer's entry.
+		log.Printf("[bus] skipping registry seq update (lock unavailable): %v", err)
+		return
+	}
+	defer lock.Release()
+
 	registry := m.loadRegistry()
 	for i, entry := range registry {
 		if entry.BusID == busID {
--- a/bus_session.go
+++ b/bus_session.go
@@ -221,6 +221,20 @@ func (m *busSessionManager) loadRegistry() []busRegistryEntry {
 }
 
 // saveRegistry writes the bus registry to disk.
+//
+// The write is atomic (tmp + rename): a plain truncate-before-write left
+// registry.json empty if the process was killed mid-write, and loadRegistry
+// swallows the resulting parse error and returns an empty registry — so the
+// next save from any process would discard every registered bus, not just the
+// one being added. Mirrors the daemon-side BusSessionManager.saveRegistry
+// (internal/engine/bus_session.go).
+//
+// NOTE: this closes the truncation→total-loss failure mode but does NOT add a
+// cross-process advisory lock. Full serialization of this root-CLI writer
+// against the daemon's writer of the same registry.json requires both sides to
+// take a shared filelock; that coordination is deferred to the L1 disposition
+// of the root bus-CLI (whether this cross-process scenario survives at all
+// depends on relocate-vs-delete). Tracked separately; see PR body.
 func (m *busSessionManager) saveRegistry(entries []busRegistryEntry) error {
 	if err := os.MkdirAll(m.busesDir(), 0755); err != nil {
 		return err
@@ -229,7 +243,15 @@ func (m *busSessionManager) saveRegistry(entries []busRegistryEntry) error {
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(m.registryPath(), data, 0644)
+	tmp := m.registryPath() + ".tmp"
+	if err := os.WriteFile(tmp, data, 0644); err != nil {
+		return err
+	}
+	if err := os.Rename(tmp, m.registryPath()); err != nil {
+		_ = os.Remove(tmp)
+		return err
+	}
+	return nil
 }
 
 // appendBusEvent appends a new CogBlock to a bus's event chain.
--- a/bus_session.go
+++ b/bus_session.go
@@ -180,8 +180,14 @@ func (m *busSessionManager) createMCPBus(sessionID, origin string) (string, erro
 }
 
 // registryLockTimeout bounds how long a registry writer waits for the
-// cross-process advisory lock before failing the operation.
+// cross-process advisory lock before failing the operation (registration:
+// correctness-critical, cold path).
 const registryLockTimeout = 5 * time.Second
+
+// registrySeqLockTimeout bounds the best-effort seq-metadata update on the
+// append hot path — short, because a skipped update self-heals on the next
+// append and a long wait would delay the append caller.
+const registrySeqLockTimeout = 500 * time.Millisecond
 
 // acquireRegistryLock takes the cross-process advisory lock guarding the
 // load-modify-save cycle on registry.json. The daemon's
@@ -332,15 +338,16 @@ func (m *busSessionManager) appendBusEvent(busID, eventType, from string, payloa
 	}
 	f.Close()
 
-	// Update registry
-	m.updateRegistrySeq(busID, newSeq, evt.Ts)
-
 	// Snapshot handlers while locked, then release BEFORE dispatching.
 	// This prevents deadlocks when handlers call appendBusEvent (e.g. tool
 	// router posting tool.result in response to tool.invoke).
 	handlers := make([]busEventHandler, len(m.eventHandlers))
 	copy(handlers, m.eventHandlers)
 	m.mu.Unlock()
+
+	// Registry seq update runs OUTSIDE m.mu: it blocks (briefly) on the
+	// cross-process filelock, and must never stall unrelated bus operations.
+	m.updateRegistrySeq(busID, newSeq, evt.Ts)
 
 	// Dispatch handlers outside the lock — safe because each handler
 	// that needs to write back to the bus will re-acquire the lock via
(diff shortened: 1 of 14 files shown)


chazmaniandinkle:
## Scope note for operator review

cog-review's latest pass (head `9d0aa2e`) confirms the actual diff is correct and well-tested, but requests expanding scope to fix two additional multi-writer races it found in unrelated files:

- `pkg/substrate/reconcile/state.go:41` (`WriteState`/`LoadState`, the generic reconcile framework backing every other resource-provider type)
- `internal/eval/mcp_tools.go:344` (`writeDispatchTrigger`/`readAndClearDispatchTriggers`)

Both are real, same-bug-class findings (the reviewer traced live daemon-vs-CLI reachability for each). However, they are **out of scope for this PR's lane** (ledger item L10, closing issue #449), which is explicitly scoped to:

1. `internal/conversations/index.go:405` (`_meta.json` writer) — fixed, this PR.
2. Audit `sdk/thread.go:250` for the same interleaving — audited, documented inline, no fix needed (no live caller).

`state.go`/`mcp_tools.go` are not named in the L10 ledger entry. `pin.go` (also flagged as an unverified note) is referenced in the ledger under a *different* item (L8, the reconcile-rescan perf class), not L10. Per the ledger's own class taxonomy, these look like either:
- new `BUG` family entries (a sibling to L10, same shape: cross-process filelock + atomic write), or
- additions to L10's scope if the operator wants this PR to absorb them.

I'm stopping reconciliation here rather than silently expanding this PR's diff into `pkg/substrate/reconcile/` and `internal/eval/` — those are outside my assigned lane's file boundaries and I don't want to risk colliding with another session's in-flight work on adjacent subsystems, or unilaterally deciding a scope question that belongs to the operator/ledger process.

**Requesting operator disposition:** either (a) confirm this PR's scope is L10-as-ledgered and merge as-is once CI is otherwise green, with the two new findings filed as follow-up ledger items, or (b) direct that this PR's scope expand to cover them.

All 6 confirmed mechanical/correctness findings from rounds 1-6 of review on this PR were fixed and are covered by regression tests (see commit history and PR description). This is the first round where the finding is a scope question rather than a defect in the diff.

chazmaniandinkle:
Final-round disclosure (per the reviewer's unverified note): `pkg/substrate/reconcile/meta.go:311` reconcileResource runs the same LoadState→WriteState cycle without AcquireStateLock. Confirmed dead code today (RunMeta/RunMetaReconcile have zero non-test callers; the `cog reconcile` CLI entrypoint was removed per the file's own header). Deliberately NOT patched here: it is already inventoried in the v1.0 alignment ledger's dead-code disposition (L7) and root-tree cascade (L1) — if it is ever revived, the state.go doc contract added by this PR requires the lock, and the conformance sweep will catch a violation. Disclosed here so the record of what was and wasn't fixed stays complete.

